PR #41, opened 2021-07-15
Multi-trenches in one namespace support

LionelJouin:
- Separation of namespace and trench name
- NSM dependencies updated to 1.0
- Targets can connect only in trenches in the same namespace


estinfra (commented):
FOSS Analysis is done. Please check the [log](https://central.jenkins.est.tech/job/est-nsm-meridio-merge-foss-analysis-master/19/consoleFull).

PR #59, opened 2021-08-27
Ambassador API documentation + vlan forwarder fix

LionelJouin:
- Ambassador API documentation 
- The label forwarder was missing in the target request.
- The kernel forwarder in the NSM helm chart has been renamed to forwarder-vlan
- Fix nfqlb index:
  - The minimum value of the index in nfqlb is 0 (instead of 1) and the maximum is 99 (instead of 100).
  - The identifier is now equal to the index - 1 (index in nfqlb). The fwmark and identifer have not changed, the min is still 1 and the max 100.

### DIFF
--- a/api/target/target.proto
+++ b/api/target/target.proto
@@ -21,25 +21,53 @@ option go_package = "github.com/nordix/meridio/api/target";
 import "google/protobuf/empty.proto";
 
 service Ambassador {
+    // Connect to a conduit, so a new interface will be created.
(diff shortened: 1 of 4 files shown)


estinfra (commented):
FOSS Analysis is done. Please check the [log](https://central.jenkins.est.tech/job/est-nsm-meridio-merge-foss-analysis-master/31/consoleFull).

PR #61, opened 2021-08-27
proxy; NSC connection to LB is renewed

zolug:
Added refresh NetworkServiceClient chain element to proxy
NSC, thus the connection gets automatically renewed before
it would expire.


estinfra (commented):
FOSS Analysis is done. Please check the [log](https://central.jenkins.est.tech/job/est-nsm-meridio-merge-foss-analysis-master/30/consoleFull).

Issue #22, opened 2021-06-15
Handling of hashing collisions


johlj:
Mitigate or manage target identifier hashing collisions

PR #67, opened 2021-09-07
Target collision fix

LionelJouin:
Adding a new status to the target gives the possibility to have an extra step in the target registration process.

The steps for the ambassador to register the target:
1. Fetching the identifiers in the NSP Service
2. Generating a new identifier based on the ones in use
3. Registering the target to the NSP service with the status as "disabled"
4. Fetching the identifiers in the NSP Service and checking if there is a collision
5. If there is: Back to step 2
5. If there is no: Updating the target status in the NSP service as "enabled"

The load-balancer will configure only the targets which are enabled, the other ones will be ignored or removed if they were existing in the LB.

Fix: #22


tedlean:
It seems fine to me reading your description. At first I wondered if the nsp could take this task. but then I realized that it is better the nsp do not include any state for this.

zolug (approved):
Seems to work fine.

estinfra (commented):
FOSS Analysis is done. Please check the [log](https://central.jenkins.est.tech/job/est-nsm-meridio-merge-foss-analysis-master/36/consoleFull).

Issue #64, opened 2021-08-31
Redundancy handling for the ipam/nsp pods


johlj:
Investigate and add high availability for the ipam/nsp pods

LionelJouin:
https://raft.github.io/

Issue #15, opened 2021-05-12
Minimize the privileges needed for running Meridio


johlj:
In order to minimize the requirements for running Meridio and strengthen security, evaluate and minimize the need for elevated privileges in Meridio

LionelJouin:
- [x] Replace `privileged: true` with specific capabilities
- [ ] Run as non root

```
securityContext:
  runAsNonRoot: true
  capabilities:
    add: ["NET_ADMIN"]
```
returns:
`Error: container has runAsNonRoot and image will run as root`

zolug:
https://github.com/Nordix/Meridio/pull/244

PR #74, opened 2021-09-17
API improvements 

LionelJouin:
_**Configuration Manager (configurationmanager.proto):**_
This service aims to replace the configmap watcher which is currently in use in most of the pods. Only the NSP service will now watch for the changes in the configmap and inform the corresponding subscribers via the GRPC stream returned by the watch functions. The subscribers can watch all different resources of Meridio (Trench, conduit, streams...) for any event (Creation, Deletion, Update).

A subscriber has the possibility to watch a specific resource based on its attributes. For instance, a stream handled by the LB, could watch the corresponding flows via the "WatchFlow" function and by specifying the stream the flows should belongs to:
```
WatchFlow(Flow{
    Stream: Stream{
        Name: "stream-1",
        Conduit: Conduit{
            Name: "load-balancer",
            Tremch: Tremch{
                Name: "trench-a",
            }
        }
    }
})
```

_**Target Registry (targetregistry.proto):**_
This service has been renamed from "NetworkServicePlateformService" to "TargetRegistry" and is now containing the attribute "Stream" (containing the name of the stream, the conduit and trench it belongs to) to specify to which stream the target is connected to. 

_**Ambassador (target.proto):**_
The function "Request" has been renamed "Open".


LionelJouin:
**_Pros:_**
- With this API, the interface is already in place to propagate the configuration. Each resource will have the responsibility to watch their own sub-resources events and create/update/delete them accordingly. For instance, a load-balancer instance (a stream) will watch each flow that belongs to it using the WatchFlow function. On creation of a flow, the flow will watch each VIP that belongs to it using the WatchVIP function. This would is be possible on each pod without the NSP Service but, as currently, for each configmap modification, an event will be triggered on every pod and each pod will have to filter the relevant events. 

- The role to watch configmaps will be assigned only to the NSP deployment, so it will no longer be needed for the target pods, which simplifies the usage of Meridio. A role and a role-binding (bound to the target SA) is currently needed for the target in order for the ambassador to watch the configmap.

**_Cons:_**
- The NSP has more responsibilities.
- On the call of a "watch" function, the NSP will return the existing resources in a "created" status. If a component gets disconnected from the NSP service, it will have to reconnect and have a "recovery" function which will detect what resources have been created/updated/removed which can be a bit tricky. 
Possible solution for this issue:
  1. The "watch" function only returns the events and "Get" are added in the API, so on a re-connection, the component can use the "Get" function to compare the local configuration with the expected one from the NSP.

tedlean:
Hi Lionel. 
I agree that it seems beneficial to collect and generalize the filtering of configuration data in a common service and this way avoid that all nVIP resources have to include this for picking configuration data exactly for their needs. 

My concern around this could be that we just create another bottleneck in the system. Instead of many pods have to interrogate the K8s API to read changes in the configmaps, there now will be placed many "watch'es" for the NSP to traverse each time a change happens in the configmaps.

Would it be possible to spread the "watch" service across more NSP instances  if this shows to be a problem? 

To me this should be doable as the "watch" service only convey information from the configmaps and do not store any state besides the "watch'es". The watch consumer should always be prepared to renew the watches in case of failure.

Another thing; would it make sense to only provide the trigger by the watch service and then let each resource read the changed configuration data directly from the configmap as of now? I know this approach will not remove the role-binding issue.

LionelJouin:
This is not currently possible since the NSP also has to support the target registry service which stores the data, for now, only in the memory. So if we deploy multiple NSP instances, the targets will be registered in only one NSP pod. I think once #64 will be solved, yes, this would be possible without any new implementation on the Configuration Manager Service.

Yes, you are right, this is possible since the components will not know what is behind the interface/service they are using to watch the configuration. We can implement this communication with the NSP service, but we could also implement the configmap watcher directly in the component (with the role/role-binding).

tedlean:
Regarding my latter proposal: My idea is to split the work to offload the NSP a bit. NSP should only provide an indication that something has happened with the watch'ed entity. Then the component should fetch the details directly from the configmap. 
This way each component only have to read the configmap when there is a change of interest and the NSP could be offloaded from returning actual configuration data. 

This is just an idea of mine, trying to distribute the load a bit better, but maybe the gain is too small?

LionelJouin:
After some discussion with @zolug, the event types have been removed from the API, so now, the components can still watch the resources with the query parameter but, now, on each event the full set of resources will be returned. In the previous version, only the resource concerned by the event was returned with an event type (create/update/delete). In the new version, it means the subscriber of the events will have to determine by itself what has been changed in the resources and update its configuration accordingly. The reason for this, is the difficulty to keep track of the events in the subscribers, the subscriber will need to update its configuration in the exact same sequence as the events arrive. 

@zolug you were mentioning an Agent which will run as a daemonset or as a deployment with multiple replicas, but I guess we could still run it in the NSP? (multiple replicas should be supported with #64).

LionelJouin:
related issues: #80 #81 #15

estinfra (commented):
FOSS Analysis is done. Please check the [log](https://central.jenkins.est.tech/job/est-nsm-meridio-merge-foss-analysis-master/41/consoleFull).

PR #85, opened 2021-10-04
fe; fix external connectivity condition check

zolug:
When either IPv4 or IPv6 connectivity is not configured,
but external connectivity is OK, then announce respective
frontend to NSP.


estinfra (commented):
FOSS Analysis is done. Please check the [log](https://central.jenkins.est.tech/job/est-nsm-meridio-merge-foss-analysis-master/42/consoleFull).

Issue #89, opened 2021-10-13
Target identifier collision between streams


LionelJouin:
Currently it is possible 2 targets open a stream (the streams have to be different for this to happen) with the same identifier which will create a conflict on the LB side since the IP rules/routes will collide.

3 solutions are possible:
1. Have common identifier range with a unique identifier for each target
  - The number of identifier in use in equal to the number of target in the trench
  - Stream and target needs to be detached in the LB. The conduit configures the target rules/routes.
2. Have common identifier range with a unique identifier for each stream opened
  - The number of identifier in use in equal to the number of stream opened in the trench
3. Separated identifier ranges
  - Each stream should have a max size + offset
  - Needs extra parameters in the configmap and some work on the Operator to find available offset
 
related issue: #32

LionelJouin:
Fixed in #91 using solution 2

Issue #31, opened 2021-07-08
Meridio traffic flows support


johlj:

tedlean:
The support for flows in for the most to be included in the FE function block.  Incoming traffic is to be classified into flows matching preconfigured 5-tuples.  Traffic that do not match any flows should be dropped silently. (We might consider adding a PM counter for dropped packages at some point. Maybe std. Linux counters will do.)
The flows are configuration-wise mapped to a stream which again is lead into a conduit.  The conduit reflects a (physical) traffic path through the cluster where the traffic is to pass a chain of network services.

PR #91, opened 2021-10-18
Flow support

LionelJouin:
Vips replaced by flow in streams on the LB
2nd solution of #89 implemented
Target example:
- ctraffic updated to latest version
- 3 TCP instances (5000, 5001, 5002) and 2 UDP instances (5003, 5004)

Example of iptables from a flow:
Flow: (stream-a via the nfqueue-num 1 ; vip1: 20.0.0.1/32, vip3: 40.0.0.0/24)
```
- name: flow-c
  vips:
    - vip1
    - vip3
  source-subnets:
    - 179.10.10.0/24
    - 180.10.10.0/24
  destination-port-ranges:
    - 7001
    - 7002
  source-port-ranges:
    - 3000-65535
    - 2000
  protocols:
    - tcp
  stream: stream-a
```

iptables:
```
-A PREROUTING -s 179.10.10.0/24 -d 20.0.0.1/32 -p tcp -m tcp --dport 7001 -m multiport --sports 3000:65535,2000 -j NFQUEUE --queue-num 1
-A PREROUTING -s 179.10.10.0/24 -d 40.0.0.0/24 -p tcp -m tcp --dport 7001 -m multiport --sports 3000:65535,2000 -j NFQUEUE --queue-num 1
-A PREROUTING -s 180.10.10.0/24 -d 20.0.0.1/32 -p tcp -m tcp --dport 7001 -m multiport --sports 3000:65535,2000 -j NFQUEUE --queue-num 1
-A PREROUTING -s 180.10.10.0/24 -d 40.0.0.0/24 -p tcp -m tcp --dport 7001 -m multiport --sports 3000:65535,2000 -j NFQUEUE --queue-num 1
-A PREROUTING -s 179.10.10.0/24 -d 20.0.0.1/32 -p tcp -m tcp --dport 7002 -m multiport --sports 3000:65535,2000 -j NFQUEUE --queue-num 1
-A PREROUTING -s 179.10.10.0/24 -d 40.0.0.0/24 -p tcp -m tcp --dport 7002 -m multiport --sports 3000:65535,2000 -j NFQUEUE --queue-num 1
-A PREROUTING -s 180.10.10.0/24 -d 20.0.0.1/32 -p tcp -m tcp --dport 7002 -m multiport --sports 3000:65535,2000 -j NFQUEUE --queue-num 1
-A PREROUTING -s 180.10.10.0/24 -d 40.0.0.0/24 -p tcp -m tcp --dport 7002 -m multiport --sports 3000:65535,2000 -j NFQUEUE --queue-num 1
```

Related issues: #31 #89


pkg/loadbalancer/flow/flow.go
```diff
@@ -50,6 +53,7 @@ func New(flow *nspAPI.Flow, nfqueueNumber int, nfqueueFactory networking.NFQueue
 
 func (f *Flow) Update(flow *nspAPI.Flow) error {
```
zolug, line 54: Seems priority update is not supported. Is it intentional?

pkg/kernel/nfqueue.go
```diff
@@ -131,139 +152,315 @@ func (nfq *NFQueue) Update(protocols []string, sourceIPs []string, destinationIP
...
-	var errFinal error
-	err := execCommand(fmt.Sprintf("nft delete chain inet %s %s", tableName, nfq.name))
-	if err != nil {
-		errFinal = fmt.Errorf("%w; %v", errFinal, err)
-	}
-	err = execCommand(fmt.Sprintf("nft delete set inet %s %s", tableName, nfq.getFullSetName(ipv4DestinationIPSet)))
-	if err != nil {
-		errFinal = fmt.Errorf("%w; %v", errFinal, err)
+	conn := &nftables.Conn{}
+	conn.DelChain(nfq.chain)
```
zolug, line 156: What about the rules? Was Flow removal tested?
https://wiki.nftables.org/wiki-nftables/index.php/Configuring_chains#Deleting_chains
"The only condition is that the chain you want to delete needs to be empty, otherwise the kernel will complain that the chain is still in use."

pkg/kernel/nfqueue.go
```diff
@@ -131,139 +152,315 @@ func (nfq *NFQueue) Update(protocols []string, sourceIPs []string, destinationIP
...
+
+func (nfq *NFQueue) configureRules() error {
+	conn := &nftables.Conn{}
+	// nft add rule inet meridio-nfqlb flow-a meta l4proto @flow-a-protocols ip saddr @flow-a-saddrs-v4 ip daddr @flow-a-daddrs-v4 th dport @flow-a-dports th sport @flow-a-sports counter queue num 1
+	ipv4Rule := &nftables.Rule{
+		Table: nfq.table,
+		Chain: nfq.chain,
+		Exprs: []expr.Any{
+			// [ meta load l4proto => reg 1 ]
+			&expr.Meta{
```
zolug, line 264: "Multiple expressions are linearly evaluated from left to right: if the first expression matches, then the next expression is evaluated and so on."
Probably performance wise it'd be better checking first if the packet is IPv4 (instead the L4 protocol type)

build/load-balancer/Dockerfile
```diff
@@ -36,7 +36,7 @@ FROM ubuntu:20.10
 RUN apt-get update -y
 run apt-get install -y libmnl-dev
 run apt-get install -y libnetfilter-queue-dev
-RUN apt-get install -y iproute2 tcpdump iptables net-tools iputils-ping ipvsadm netcat
+RUN apt-get install -y iproute2 tcpdump iptables net-tools iputils-ping ipvsadm netcat ipset nftables
```
zolug, line 39: is ipset needed? nftables sets are not ipsets

pkg/loadbalancer/flow/flow.go
LionelJouin: Yes, it is intentional. To update it, it would require to delete the chain and add it again, so it is not possible without traffic interruption.

pkg/kernel/nfqueue.go
LionelJouin: Right, I tried it without removing the rules, and it worked fine, traffic was no longer working, and I could create the same flow again after removing it

build/load-balancer/Dockerfile
LionelJouin: No ipset is not needed, I added it when I was working on the version with iptables + ipset

pkg/kernel/nfqueue.go
LionelJouin: Do you mean inverting:
```
 meta load l4proto => reg 1 ]
[ lookup reg 1 set flow-a-protocols ]
```
with:
```
[ meta load nfproto => reg 1 ]
[ cmp eq reg 1 0x00000002 ]
```
I just took these 2 commands:
```
nft add rule inet meridio-nfqlb flow-a meta l4proto @flow-a-protocols ip saddr @flow-a-saddrs-v4 ip daddr @flow-a-daddrs-v4 th dport @flow-a-dports th sport @flow-a-sports counter queue num 1
nft add rule inet meridio-nfqlb flow-a meta l4proto @flow-a-protocols ip6 saddr @flow-a-saddrs-v6 ip6 daddr @flow-a-daddrs-v6 th dport @flow-a-dports th sport @flow-a-sports counter queue num 1
```
and used "--debug all" which returns all expressions, and I converted it to the code.

zolug: Yes, that what i meant, to have the IP version check first

pkg/loadbalancer/flow/flow.go
zolug: I was more curious about the requirement. But i'd be fine with the approach that one needs to delete a Flow on the operator side first if wishes to change the prio.

pkg/kernel/nfqueue.go
zolug: That would mean, that the documentation is misleading.
Could you please double check whether during Flow removal an error is received in the code upon deleting the chain? And check how the nftables looks like after a Flow removal?

LionelJouin: I will check again if there any error. Otherwise I tried this morning again, and I checked the nftables, everything that belongs to the flow was removed correctly (no more set, no more rule, no more chain). And I triied to create it again, everything was fine (no duplicates or anything)

LionelJouin: I just tried, and no, I don't see any error on flow removal

zolug (approved):

estinfra (commented):
FOSS Introduction is done. Please check the [logs](https://artifactory.est.tech/artifactory/proj-est-nsm-generic-local/reports/meridio/master/07f2686).

PR #102, opened 2021-11-04
cm reader: omitempty for BFD/BGP specific params when undefined

mandydydy:
They will shown to be 0 when undefined, which is confusing


estinfra (commented):
FOSS Introduction is done. Please check the [logs](https://artifactory.est.tech/artifactory/proj-est-nsm-generic-local/reports/meridio/master/1cd31e1).

PR #103, opened 2021-11-07
ovl/meridio; document load of private registry

uablrek:
Had to find out what images to load. Good to save time for others using `private-reg`.

Also I prefer to be able to see the actual manifests and apply them on target.

### DIFF
--- a/docs/demo/deployments/xcluster/ovl/meridio/README.md
+++ b/docs/demo/deployments/xcluster/ovl/meridio/README.md
@@ -31,3 +31,24 @@ issues, where usually one of the VPP forwarder gets killed by the OS.
 
 Refer to [Meridio description](https://github.com/Nordix/Meridio/blob/master/docs/demo/xcluster.md) for details. 
 
+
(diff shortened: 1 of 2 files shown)


docs/demo/deployments/xcluster/ovl/meridio/README.md
```diff
@@ -31,3 +31,24 @@ issues, where usually one of the VPP forwarder gets killed by the OS.
...
+## Private registry and local manifests
+
+Generate local manifests in the "default/" directory and check images
+missing in the `private-reg`;
+
+```
+mdir=$(readlink -f ../../../../../..)
+dst=./default/etc/kubernetes/meridio
+mkdir -p $dst
+for n in spire nsm nsm-vlan; do
```
zolug, line 44: nsm is not used anymore, nsm-vlan is the one to go with
(at least until vlan forwarder functionality gets merged into vpp)

uablrek:
Obsolete

PR #105, opened 2021-11-09
FE; parse and apply BFD attributes

zolug:
closes: #100 

-Min tx/rx intervals and multiplier are fetched from the config
 generated by the Operator (as part of the gateway settings)
-Requires at least BIRD 2.0.8
-BGP: BFD options are set per peer i.e. per gateway
-Static+BFD: due to limitations in BIRD they are applied per
 external interface (meaning only 1 BFD configuration can
 take effect per Frontend)

-BFD session state of peer IP is ignored for Static protocol
 if BFD is not configured
-alpine:edge base image


pkg/configuration/reader/converter.go
```diff
@@ -163,15 +163,25 @@ func ConvertGateways(gateways []*Gateway, trench *nspAPI.Trench) []*nspAPI.Gatew
...
-			Protocol:   gateway.Protocol,
 			HoldTime:   uint32(gateway.HoldTime),
-			Trench:     trench,
+			Bfd:        gateway.BFD,
+			BfdSpec: func() *nspAPI.Gateway_BfdSpec {
+				if !gateway.BFD {
+					return nil
+				}
+				return &nspAPI.Gateway_BfdSpec{
+					MinTx:      uint32(gateway.MinTx),
```
LionelJouin, line 179: Maybe MinTx, MinRx, Multiplier and HoldTime should be set to uint32 in /pkg/configuration/reader/model.go (this change would require some modifications in the operator)

Some gateway fields could be empty in the configmap, would default values be needed?

LionelJouin (approved):

pkg/configuration/reader/converter.go
zolug: In case of BFD parameters there are default values on the Operator side for both BGP and Static protocols. So they will never be empty.
That's needed because in BIRD BFD default values are different than the ones we intend to use.

On the other hand, if any of the BFD attributes took the variables default value, it should not cause any problem, as the FE will ignore that, thus BIRD's default setting for that attribute will take effect.

estinfra (commented):
FOSS Introduction is done. Please check the [logs](https://artifactory.est.tech/artifactory/proj-est-nsm-generic-local/reports/meridio/master/a2a32d2).

PR #115, opened 2021-11-18
lb; enable IP defrag to support Stream/Flow selection

zolug:
-Load kernel's IP defragmentation via conntrack allowing Flow rules
to match L4 information (e.g. ports).
-Do not allow defragmentation of IP packets arriving from the direction
of targets. Thus outbound IPv4 packets leaving the LB could keep their
originating source's "PMTU information".
-Forbid conntrack to do "book-keeping" in order to not "litter" memory.

### DIFF
--- a/pkg/loadbalancer/stream/loadbalancer.go
+++ b/pkg/loadbalancer/stream/loadbalancer.go
@@ -46,6 +46,7 @@ type LoadBalancer struct {
 	ctx                        context.Context
 	cancel                     context.CancelFunc
 	pendingTargets             map[int]types.Target // key: Identifier
+	defrag                     *Defrag
 }
 
 func New(stream *nspAPI.Stream, targetRegistryClient nspAPI.TargetRegistryClient, configurationManagerClient nspAPI.ConfigurationManagerClient, m int, n int, nfqueue int, netUtils networking.Utils) (types.Stream, error) {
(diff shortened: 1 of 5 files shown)


zolug:
closes: #116

pkg/loadbalancer/stream/loadbalancer.go
```diff
@@ -64,6 +65,13 @@ func New(stream *nspAPI.Stream, targetRegistryClient nspAPI.TargetRegistryClient
 		nfqueue:                    nfqueue,
 		pendingTargets:             make(map[int]types.Target),
 	}
+	// first enable kernel's IP defrag except for the interfaces facing targets
+	// (defrag is needed by Flows to match rules with L4 information)
+	loadBalancer.defrag, err = NewDefrag(types.InterfaceNamePrefix)
```
LionelJouin, line 70: Having NewDefrag here means the table, chains and rules for defrag will be created each time a stream is created (so multiple times if there is multiple streams). Was this the expected behavior?

zolug: Yes, I checked and table/chain create will not return an error if the table/chain is already present, and creates what's been asked only if it does not exist.

LionelJouin: ok, otherwise this could be moved directly in cmd/load-balancer

zolug: keeping the code at its original place for now until multiple conduits will get introduced

estinfra (commented):
FOSS Introduction is done. Please check the [logs](https://artifactory.est.tech/artifactory/proj-est-nsm-generic-local/reports/meridio/master/47aed37).

PR #127, opened 2021-12-01
NSP/IPAM data persistence

LionelJouin:
Simplified watcher pattern (for target registry, other watchers pattern should follow the same design)
New TargetRegistry Interface
Target identifier collision fix (targets were checking only identifier of disabled targets instead of any status)

CGO enabled in the NSP compilation to allow SQLite to work
Sqlite Registry saving the db in a persistent volume
NSP changed to statefulset with a volume claim

go-ipam dependency removed
New IPAM implementation:
- Storage support (SQLite and memory)
- Allocate single IP or subnet from subnet pool
IPAM changed to statefulset with a volume claim
IPAM server stores data in SQLite
GCO Enabled on IPAM to support SQLite

#64


estinfra (commented):
FOSS Introduction is done. Please check the [logs](https://artifactory.est.tech/artifactory/proj-est-nsm-generic-local/reports/meridio/master/30e9999).

Issue #104, opened 2021-11-08
Refactoring and stability


LionelJouin:
1. NSM
- [ ] Refactor /pkg/client to use NSM APIs (so with context.WithTimeout)
- [ ] Refactor /pkg/nsm
- [ ] Refactor or remove /pkg/nse
- [ ] Replace InterfaceMonitor
- [ ] Split NSM pkg and Spire (Spire pkg to be used in Meridio with the IPAM/NSP APIs)
2. Load balancer
- [ ] NewSimpleNetworkService to a new proper conduit handler struct
3. Ambassador
- [ ] Replace config watcher with the NSP API
- [ ] Rename target by Ambassador (pkg, cmd, container (Docker, CI, Helm))
4. Proxy
- [ ] Multi conduit support to be able to use any conduit name (currently conduit name is static since it is a env variable in the conduit)
- [ ] Replace config watcher with the NSP API
5. Internal APIs (IPAM, NSP, Ambassador)
- [ ] Usage of context.WithTimeout
- [ ] Retry mechanism
6. IPAM
- [ ] API re-work + versioning?
- [ ] Replace dependency (metal-stack/go-ipam)
7. /pkg/networking
- [ ] re-work interfaces
8. configuration (env variables)
- [ ] replace NSM prefix with MERIDIO
- [ ] replace dependency? 
9. Move all example in a single directly (/examples/target)
- [ ] New go module
- [ ] /cmd/ctraffic ; /build/ctraffic ; /test/applications -> /examples/target
10. other
- [ ] recover mechanism on container restart 
- [ ] keep alive mechanism (target registered but still alive?, proxy releasing subnets to IPAM...)
- [ ] check goroutine leaks
- [ ] drain channels if event sent multiple times
- [ ] modify linters enabled

PR #139, opened 2022-01-12
Privileges reduced

LionelJouin:
Old code removed and replaced with the configuration manager NSP service
Common helm chart for targets removed
New env variables required on the proxy and target
Service account removed (LB and Proxy)
Capabilities set to NET_ADMIN

#15 #104


zolug (approved):

cmd/proxy/main.go
```diff
@@ -133,20 +137,41 @@ func main() {
 	defer ep.Delete()
 	probe.CreateAndRunGRPCHealthProbe(ctx, health.NSMEndpointSvc, probe.WithAddress(ep.GetUrl()), probe.WithSpiffe())
 
-	// TODO: use NSP based config watcher
-	configWatcher := make(chan *configuration.OperatorConfig)
-	configurationWatcher := configuration.NewOperatorWatcher(config.ConfigMapName, config.Namespace, configWatcher)
-	go configurationWatcher.Start()
 	health.SetServingStatus(ctx, health.NSPCliSvc, true)
```
zolug, line 140: This should be placed after successfully connected with NSP service.

pkg/target/conduit/configuration.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+type Configuration struct {
+	Conduit                    types.Conduit
+	ConfigurationManagerClient nspAPI.ConfigurationManagerClient
+	Cancel                     context.CancelFunc
+}
+
+type SetVips interface {
```
zolug, line 34: Is this used at all?

LionelJouin: No, I had the plan to use it to replace the "Conduit" in the Configuration struct, so it would be easier to create unit tests later with a small mock of it.

pkg/target/conduit/conduit.go
```diff
@@ -143,6 +150,7 @@ func (c *Conduit) Disconnect(ctx context.Context) error {
 	c.deleteVIPs(c.vips)
 	c.nexthops = []string{}
 	c.tableID = 1
+	c.Configuration.Delete()
```
zolug, line 153: Since watching changes in vips are done via Configuration, wouldn't it be better Deleting the Configuration before cleaning up the vips?
On the other hand that cancel based stopping of WatchVIPs is asynchronous imho. So probably it is not granted that watch method would conclude before moving on to vip clean-up part...

estinfra (commented):
FOSS Introduction is done. Please check the [logs](https://artifactory.est.tech/artifactory/proj-est-nsm-generic-local/reports/meridio/master/47943bb).

PR #145, opened 2022-01-25
Target (cmd, api, build, pkg) renamed to TAP/TAPA

LionelJouin:
- cmd/tapa
- pkg/ambassador/tap
- api/ambassador
- target example tapa container


LionelJouin:
will be merged with #152

PR #149, opened 2022-01-25
Remove common helm chart (target example)

LionelJouin:
had to remove it with: #139


estinfra (commented):
FOSS Introduction is done. Please check the [logs](https://artifactory.est.tech/artifactory/proj-est-nsm-generic-local/reports/meridio/master/865aecc).

PR #150, opened 2022-01-25
Fixed: return nil -> return err

uablrek:
Another return fix.


LionelJouin (approved):

LionelJouin:
Yes, we would also need to add the copyright header in the Operator repository

estinfra (commented):
FOSS Introduction is done. Please check the [logs](https://artifactory.est.tech/artifactory/proj-est-nsm-generic-local/reports/meridio/master/bc1ecf0).

PR #156, opened 2022-02-04
Fixed if err != nil {return nil} bugs

uablrek:
Some more.

In `RemoveConduit()` perhaps the Disconnect error should be ignored instead?


LionelJouin (approved):
With the current implementation, the user will receive the error, so he could again try to Disconnect the conduit.

uablrek:
> With the current implementation, the user will receive the error, so he could again try to Disconnect the conduit.

So, it's ok to merge?

uablrek:
Obsoleted by the tapa anbassador

PR #163, opened 2022-02-15
Retry pkg

LionelJouin:
The Retry pkg is a common pkg which can be used for all functions which could need a retry mechanism, for instance, the LB component needs to watch the target using the NSP service, but if the service goes down, the LB would need to call the NSP service again.

* Allows the configuration of the retry trigger (can be a delay, a system event...)
* Allows the configuration of the retry condition (can be a maximum of attempts, depends on the error returned by the retryable func...)
* Allows the configuration of the context

issue: #147

### DIFF
--- a/pkg/retry/retry.go
+++ b/pkg/retry/retry.go
@@ -1,0 +1,63 @@
+/*
+Copyright (c) 2021-2022 Nordix Foundation
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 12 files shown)


pkg/retry/retry.go
```diff
@@ -0,0 +1,63 @@
...
+// retry condition is fulfilled or until the context is Done.
+// The following options are supported:
+// - Retry condition: condition for retries to terminate (e.g. retryableFunc has not returned any error, max attempts...)
+// - Retry trigger: trigger to continue in the next try (e.g. Delay, an system event...)
+// - Context: context that can be used to terminate the function
+// By default, these options are applied:
+// - Retry condition: retryableFunc has not returned any error
+// - Retry trigger: 10 milliseconds delay
+// - Context: empty context
+func Do(retryableFunc RetryableFunc, options ...Option) error {
```
zolug, line 36: If it's intentional, it could make sense to highlight that there's no retry limit by default.

zolug (approved):

Issue #133, opened 2021-12-14
fix processing delay of Targets after LB scale-out to minimize traffic disturbance


zolug:
When scaling out LBs, the fresh LBs usually get informed about the existing Targets through NSP API before the NSM interfaces connecting it with the Proxies could come up.
A Target in this case can not be configured and will get appended to a so called pending queue by the respective Stream. Currently processing of the pending queue is attempted every 10 seconds, which might lead to traffic disturbance (as inbound packets might arrive to the LB once the external connectivity is established).

One possible fix could be to rely on interface creation events in LB (SimpleNetworkService:InterfaceCreated()) and poke the Streams so that they could attempt draining their pending Target queues.

PR #171, opened 2022-02-18
lb; process pending Targets when new NSM interface appears

zolug:
closes: #133

### DIFF
--- a/pkg/loadbalancer/stream/loadbalancer.go
+++ b/pkg/loadbalancer/stream/loadbalancer.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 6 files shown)


pkg/loadbalancer/stream/loadbalancer.go
```diff
@@ -398,3 +426,21 @@ func (lb *LoadBalancer) removeFromPendingTarget(target types.Target) {
...
+		return
+	}
+	lb.mu.Unlock()
+	// send trigger if interface is of interest (i.e. lb NSE)
+	if strings.HasPrefix(intf.GetName(), GetInterfaceNamePrefix()) {
+		lb.pendingCh <- struct{}{}
+	}
+}
+
+func (lb *LoadBalancer) InterfaceDeleted(intf networking.Iface) {
```
LionelJouin, line 445: This would also need to be implemented, when an interface is removed, the routes of some targets might be removed. The verifyTargets has for responsibility to check that, clean them and add them back to the pending target list.

LionelJouin (approved):

Issue #174, opened 2022-02-23
Unique identifier range per stream 


LionelJouin:
Motivations:
1. The stream identifiers will be clearly separated, so they can handle more target without impact on each other.
2. Some stream (e.g. SCTP ones) could have lower range to save resources (memory). 

- [x] Configmap + NSP API changes
- [x] Load-balancer: set correct identifier range for each stream
- [x] TAPA: generate within the stream identifier range 
- [x] Ambassador: Calculate and add a unique range while adding the streams in the configmap 
- [x] Ambassador: Env variable setting the default range size 

Related issue: https://github.com/Nordix/Meridio/issues/167

PR #175, opened 2022-02-25
Identifier range per stream

LionelJouin:
## Description
API:
- New max-targets property in Configmap and NSP API Stream

LB:
- NFQLB pkg modified to accept target index and identifier
- M, N Constants removed and modified with N = max-targets, and M = max-targets * 100

TAP ambassador:
- Identifier generated based only on identifiers registered in the same stream (previously based on identifier registered in the same conduit)
- NSP Stream (containing the max-targets) required in parameter to open a tapa stream

There is no main modification in the unit tests, it's only adaptation. 

Future improvement: restricted tables id in the identifier offset generator

_**Changes in the Operator:**_
the max-targets property should be set in the configmap.

## Issue link
https://github.com/Nordix/Meridio/issues/174

## Checklist

- Purpose
    - [ ] Bug fix
    - [x] New functionality
    - [ ] Documentation
    - [ ] Refactoring
    - [ ] CI
- Test
    - [x] Unit test
    - [ ] E2E Test
    - [x] Tested manually
- Introduce a breaking change
    - [x] No
    - [ ] No


uablrek:
General comment; IMHO (start, count) is better than (start, end). E.g (start=1, count=99) == (start=1, end=100)

There is _always_ confusion if the "end" is inclusive or not.

LionelJouin:
In the NSP API proto file it is specified the end is inclusive: https://github.com/Nordix/Meridio/blob/fac144150409c3160c3d89c3b817dfd082f34347/api/nsp/v1/model.proto#L40
(so in the case where the range is s(tart=1, end=100), the size would be 100)

uablrek:
I rest my case :smile:

pkg/configuration/reader/model.go
```diff
@@ -34,8 +34,9 @@ type StreamList struct {
 }
 
 type Stream struct {
-	Name    string `yaml:"name"`
-	Conduit string `yaml:"conduit"`
+	Name            string `yaml:"name"`
+	Conduit         string `yaml:"conduit"`
+	IdentifierRange *Range `yaml:"identrifier-range"`
```
uablrek: Typo; identrifier-range

pkg/configuration/reader/model.go
```diff
@@ -94,3 +95,8 @@ type Gateway struct {
 	Multiplier uint   `yaml:"multiplier,omitempty"`
 	Trench     string `yaml:"trench"`
 }
+
+type Range struct {
+	Start uint32 `yaml:"Start"`
```
uablrek: Case-mix; Start / end. start / end is better

pkg/configuration/reader/model.go
LionelJouin: fixed in "[Identifer range on LB](https://github.com/Nordix/Meridio/pull/175/commits/18700b454dbec02c9176a81bbb3c5637c9660caf)"

pkg/configuration/reader/model.go
LionelJouin: fixed in "[Identifer range on LB](https://github.com/Nordix/Meridio/pull/175/commits/18700b454dbec02c9176a81bbb3c5637c9660caf)"

pkg/loadbalancer/nfqlb/nfqlb.go
```diff
@@ -55,12 +55,12 @@ func New(name string, m int, n int, nfqueue int) (*nfqlb, error) {
 	return lb, nil
 }
 
-func (n *nfqlb) Activate(identifier int) error {
+func (n *nfqlb) Activate(index int, identifier int) error {
 	var stderr bytes.Buffer
 	cmd := exec.Command(
 		nfqlbCmd,
 		"activate",
-		fmt.Sprintf("--index=%d", identifier-1),
```
uablrek, line 63: Why not ditch this -1 thing while it's still time?

uablrek: Sorry, looks like this was already removed

uablrek:
General comment;
```
      identifier-range:
        start: 100
        end: 199
```
feels `nfqlb` specific to me. This may change, nfqlb may get a full user-space implementation without fwmarks, or be replaced.

IMHO a `max-targets` parameter would be generic. Any load-balancer would need it (unless a complicated dynamic sizing is implemented). However that would require that the "start" parameter must be dicided internally by Meridio when `nfqlb` with fwmark is used.

uablrek:
Improve resilience to misconfigurations. If the `identifier-range` is unspecified the `nsp` process crashes with a segv;
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x8a707a]

goroutine 35 [running]:
github.com/nordix/meridio/pkg/configuration/reader.ConvertStreams({0xc000010500, 0x1, 0xc000010878}, {0xc000010900, 0x1, 0xc00019a0f0})
        /home/uablrek/go/src/github.com/Nordix/Meridio/pkg/configuration/reader/converter.go:88 +0x21a
github.com/nordix/meridio/pkg/configuration/reader.ConvertAll(0xc0001a7740, {0xc000010468, 0x1, 0x1}, {0xc000010500, 0x1, _}, {0xc0000105e0, 0x1, 0x1}, ...)
        /home/uablrek/go/src/github.com/Nordix/Meridio/pkg/configuration/reader/converter.go:42 +0xfb
github.com/nordix/meridio/pkg/configuration/monitor.(*ConfigMapMonitor).eventHandler(0xc000458ff0, 0xc000270200)
        /home/uablrek/go/src/github.com/Nordix/Meridio/pkg/configuration/monitor/monitor.go:108 +0xc7
github.com/nordix/meridio/pkg/configuration/monitor.(*ConfigMapMonitor).watchEvent(0x0, 0x0)
        /home/uablrek/go/src/github.com/Nordix/Meridio/pkg/configuration/monitor/monitor.go:85 +0xa5
github.com/nordix/meridio/pkg/configuration/monitor.(*ConfigMapMonitor).Start.func1()
        /home/uablrek/go/src/github.com/Nordix/Meridio/pkg/configuration/monitor/monitor.go:69 +0x2b7
github.com/nordix/meridio/pkg/retry.Do(0xc000571f88, {0xc000547f70, 0x3, 0x0})
        /home/uablrek/go/src/github.com/Nordix/Meridio/pkg/retry/retry.go:46 +0xa8
github.com/nordix/meridio/pkg/configuration/monitor.(*ConfigMapMonitor).Start(0xc000458ff0, {0x18d6950, 0xc000046050})
        /home/uablrek/go/src/github.com/Nordix/Meridio/pkg/configuration/monitor/monitor.go:61 +0x125
created by main.main
        /home/uablrek/go/src/github.com/Nordix/Meridio/cmd/nsp/main.go:82 +0x438
```
That is not user friendly at any rate.

If end < start the load-balancer fails to create a shm. In the logs;
```
Mar 10 07:00:56.617 [ERRO] updateStreams err: %!w(<nil>); exit status 1; N can't be larger than M;       ;       
```
I think that is OK.

LionelJouin:
right, this is possible. Could it be a problem if we are not synchronizing the "start" value between the load-balancer? Like if lb-1 uses 1 as start for stream-a and 20 as start for stream-b while lb-2 uses 20 as start for stream-a and 1 as start for stream-b.

@zolug @johlj @tedlean any opinion on this proposal?

For the NSP that crashes if the identifier range is not set in the configmap, I think we could address this in another issue since it's the same problem for all properties.

uablrek:
> Could it be a problem if we are not synchronizing the "start" value between the load-balancer?

Can't see that that would be a problem. The important thing is that the same hash goes to the same target. That basically means that the same _index_ should go to the same target on all LBs. Which fwmark that is used is not relevant.

pkg/loadbalancer/nfqlb/nfqlb.go
LionelJouin: I had to re-add it, otherwise there will be an NBC. From the beginning we are generating identifier from 1 to 100 (and not 0 to 99), if we change it, then the old TAPA won't be compatible with this new PR.

uablrek:
If I start without modifying the config-map, I get;
```
/ # nfqlb show --shm=tshm-stream1
Shm: tshm-stream1
  Fw: own=0
  Maglev: M=3, N=0
   Lookup: -1 -1 -1...
   Active:
```

uablrek:
If "max-targets" is specified without this patch start fails => upgrade not possible (see also #180)

LionelJouin:
Right, I will add a default value to max-target. 

But this won't happen with the operator since the Meridio version and the Meridio operator version are sync. The operator identifier range PR will always set a value to max-target.

uablrek:
Is the operator always upgraded before for instance nsp and load-balancers? Otherwise there might be problems.

LionelJouin:
Yes, the operator is always upgraded before the other components (NSP, stateless-lb, ipam...)

PR #187, opened 2022-03-25
Add port NAT

uablrek:
A port-NAT is requested to allow un-privileged ports in application pods while exposing priviledged, like 80 -> 8080 for http.

#### Note to reviewers

The `nft` operations in a transaction (before [nftables.Flush()](https://pkg.go.dev/github.com/google/nftables#Conn.Flush)) are [atomic](https://wiki.nftables.org/wiki-nftables/index.php/Atomic_rule_replacement). Meaning that an updated address set can safely be flushed before adding the updated addresses. There is no need for a clumsy and error-prone add-new-remove-leftovers procedure.


uablrek:
I will add tests, but the code is more-or-less done.

api/nsp/v1/utils.go
```diff
@@ -135,7 +139,16 @@ func (f *Flow) DeepEquals(f2 *Flow) bool {
 
 // compareSlice -
 // Returns false if the two slices are not equal by content
+// Intentional? This is more a set-compare. a=[x,x,y], b=[y,y,x] would return true.
```
zolug, line 142: Wasn't intentional :)
However considering the usage it doesn't make a difference.

pkg/loadbalancer/flow/nfthandler.go
```diff
@@ -0,0 +1,475 @@
...
+				CsumType:       expr.CsumTypeInet,
+				CsumOffset:     16,
+			},
+			// [ counter pkts 0 bytes 0 ]
+			&expr.Counter{
+				Bytes:   0,
+				Packets: 0,
+			},
+			// [ notrack ]
+			&expr.Notrack{},
```
zolug, line 452: The chain prio is 100, but AFAIK notrack must be added before ChainPriorityConntrack to take effect

pkg/loadbalancer/flow/flow.go
```diff
@@ -17,50 +17,138 @@ limitations under the License.
...
 }
 
+// Update updates a flow
 func (f *Flow) Update(flow *nspAPI.Flow) error {
-	if f.Flow == nil || !f.Flow.DeepEquals(flow) {
-		f.Flow = flow
-		logrus.Infof("Update flow: %v", f.Flow)
-		if f.nfqueueLoadBalancer != nil {
-			return f.nfqueueLoadBalancer.SetFlow(f.Flow)
+	logrus.Infof("Flow:Update: %v -> %v", f.Flow, flow)
```
zolug, line 90: This will flood the logs. Any change leads to pushing all the flows to the lb. (Including NSP failover.)

pkg/loadbalancer/flow/nfthandler.go
```diff
@@ -0,0 +1,475 @@
...
+		Name:     "flow-" + flowName,
+		Table:    nfth.table,
+		Type:     nftables.ChainTypeFilter,
+		Hooknum:  nftables.ChainHookPrerouting,
+		Priority: nftables.ChainPriorityNATSource,
+	})
+	conn.FlushChain(ch)
+
+	ipv4Vips, ipv6Vips := nfth.vipSets(flowName)
+	for _, prot := range protocols {
```
zolug, line 115: Actually it is possible to use a single rule to match multiple l4 protocols port info to lower the overall number of rules (https://www.mankier.com/8/nft#Payload_Expressions-Raw_Payload_Expression). However I guess it does not seem plausible that a "nat flow" will be configured with multiple protocols, so it would not worth the effort.

pkg/loadbalancer/flow/flow.go
```diff
@@ -17,50 +17,138 @@ limitations under the License.
...
+	if flow.LocalPort != 0 {
+		dport, err := ensureSingleDport(flow)
+		if err != nil {
+			return err
+		}
+		oldDport, _ := ensureSingleDport(f.Flow)
+
+		if flow.LocalPort != f.Flow.LocalPort || dport != oldDport {
+			logrus.Infof("Flow:Update: Port NAT: %v -> %v", dport, flow.LocalPort)
+			err = f.nftHandler.PortNATSet(
```
zolug, line 111: What if the flow had not LocalPort before (i.e. f.Flow.LocalPort==0)?

pkg/loadbalancer/flow/flow.go
```diff
@@ -17,50 +17,138 @@ limitations under the License.
...
+				flow.Name, flow.Protocols, dport, uint(flow.LocalPort))
+			if err != nil {
+				return err
+			}
+		}
+		err = f.nftHandler.PortNATSetAddresses(flow)
+		if err != nil {
+			return err
```
zolug, line 119: is it worth to clean-up here?
f.nftHandler.PortNATDelete(f.Flow.Name)

pkg/loadbalancer/flow/nfthandler.go
uablrek: Does that apply when the purpose is not to prevent packet reassembly?

The reason here applies for *all* traffic through the Meridio LB; since return traffic may take another path the conntracker may never see the complete connection. So connection tracking is pointless.

The Meridio LB should perhaps have a general `notrack` in a high prio chain, but that is not a responsibility in this PR. I can remove Notrack, but I will not re-prioritize the chain.

pkg/loadbalancer/flow/nfthandler.go
uablrek: I let is be as-is. SCTP is not mentioned in the man page and may require special treatement anyway.

api/nsp/v1/utils.go
uablrek: I modify the comment but leave it since there may be more nitpickers than me.

pkg/loadbalancer/flow/nfthandler.go
uablrek: Left as-is for now.

pkg/loadbalancer/flow/flow.go
uablrek: Changed to trace level

pkg/loadbalancer/flow/flow.go
uablrek: Good catch. I missed this. Code and test added.

pkg/loadbalancer/flow/flow.go
uablrek: Good comment. But the more I think about it the less certain I become about the best action.

The situation is that the flow is updated either with an incorrect configuration or there is some problem with nft. What we don't want is to create a configuration that doesn't allow the user to correct things.

uablrek: I decided to *not* cleanup anything.

The localPort is != 0 and a natural recovery user action would be to delete the flow and re-create it. On delete everything will be cleaned up.

uablrek:
I have updated, rebased and removed the xcluster tests. I will att the xcluster tests in another PR (without review).

zolug (approved):

PR #188, opened 2022-03-29
NSM uplift to 1.3.0-rc.1

LionelJouin:


uablrek (approved):
Re-built images and tested.

uablrek:
Tests were made with NSM images v1.2.0, but with this PR applied. NSM v1.3.0-rc.1 seems broken.

Issue #182, opened 2022-03-16
proxy; pending NSC Request for removed NSE gets stuck


zolug:
Proxy learns and tries to establish NSM connection with new load-balancer NSEs relying on FullMeshNetworkServiceClient.

However, if a particular LB NSE disappears before the proxy could have successfully connected to it, the pending Request blocks FullMeshNetworkServiceClient from learning about NSE changes and from possibly cancelling invalid connections/requests. There's also no timeout for the Request to succeed, thus such Request gets stuck.

Upgrade might be heavily affected by this problem where a new proxy still learns about an old LB NSE to be removed.

Rework FullMeshNetworkServiceClient: NSC Request must not block new NSE registry client events.

PR #189, opened 2022-03-29
proxy; pending fullMesh client requests can be closed

zolug:
closes: #182 

The proxy functionality that processes NetwotkServiceEndpoint changes
is no longer blocked on establishing connection to new a endpoint.
Pending connection requests towards disappeared endpoints are removed.
So e.g., during Meridio upgrade the terminating NetworkServiceEndpoints
of an old release can't stall operation of freshly started proxies anymore.

-pkg/client moved to cmd/proxy/internal
-clean-up can take place upon proxy shutdown (e.g., closing fullMesh clients,
 unregistering nse)


cmd/proxy/internal/service/client.go
```diff
@@ -0,0 +1,90 @@
...
+
+	return fullMeshClient
+}
+
+func StartNSC(fullMeshClient client.NetworkServiceClient, networkServiceName string) {
+	logrus.Infof("Start Full Mesh NSC")
+	err := fullMeshClient.Request(&networkservice.NetworkServiceRequest{
+		Connection: &networkservice.Connection{
+			NetworkService: networkServiceName,
+			Labels:         map[string]string{"forwarder": "forwarder-vpp"}, // TODO: is it still needed?
```
LionelJouin, line 76: no, It can be removed, we needed it when 2 forwarders were used (vlan and vpp)

LionelJouin (approved):
cmd/proxy/internal/client/client.go only wraps the Request/Close NSM provides, so I think we could remove it. printConnectionExpTime could probably be a new client chain element. And the full mesh could become a new common pkg (so it could be re-use by NSM team for testing). These are only ideas that can be addressed later in a future PR.

Otherwise in general it looks good. I tried it manually, and proxy and LB resiliency worked fine.

PR #202, opened 2022-04-13
Improve TAPA stability

LionelJouin:
It can happen NSM blocks requests for some reason.
A static timeout has been added, so the request can be retried.
If an error was returned the message couldn't be printed, and an empty connection was set.

### DIFF
--- a/pkg/ambassador/tap/conduit/conduit.go
+++ b/pkg/ambassador/tap/conduit/conduit.go
@@ -120,7 +120,7 @@ func (c *Conduit) Connect(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	logrus.Infof("Conduit connected: %v", c.Conduit)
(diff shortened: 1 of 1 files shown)


pkg/ambassador/tap/conduit/conduit.go
```diff
@@ -120,7 +120,7 @@ func (c *Conduit) Connect(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	logrus.Infof("Conduit connected: %v", c.Conduit)
+	logrus.Infof("Conduit connected: %v ; NSM Connection: %v", c.Conduit, c.connection)
```
zolug, line 123: _c.connection_ is still nil here (either log _connection_ or move the printout)

LionelJouin: thank you, fixed

Issue #192, opened 2022-04-08
Hearthbeat function for the tapa


johlj:
If the tapa container is suddenly removed or if it can't reach the nsp during a graceful shutdown there will be leftover identifiers that still receive traffic which won't be cleared until redeployment

Implement a function for the nsp to detect stale identifiers

PR #205, opened 2022-04-22
KeepAlive Mechanism on NSP Target Registry

LionelJouin:
Documentation for the TAPA: https://github.com/Nordix/Meridio/blob/documentation/docs/tapa.md#refresh
NSP: https://github.com/Nordix/Meridio/blob/documentation/docs/nsp.md#target-registry

Requires this fix: https://github.com/Nordix/Meridio/pull/206
Issue: https://github.com/Nordix/Meridio/issues/192


cmd/frontend/internal/frontend/service.go
```diff
@@ -309,6 +317,17 @@ func (fes *FrontEndService) Monitor(ctx context.Context) error {
 						if err := announceFrontend(fes.targetRegistryClient); err != nil {
 							logrus.Infof("FrontEndService: failed to announce frontend connectivity (err: %v)", err)
 						}
+						// refresh NSP entry
+						var refreshCtx context.Context
+						refreshCtx, refreshCancel = context.WithCancel(context.TODO())
+						defer refreshCancel()
```
zolug, line 325: What happens on link flapping? Imho this leads to memory leak:
Every time when the external connectivity recovers a new deferred function is added to the stack.

I would sg like this to the end of the for loop instead:
```
if refreshCancel != nil {
	refreshCancel()
}
```

cmd/frontend/internal/frontend/service.go
```diff
@@ -253,6 +254,7 @@ func (fes *FrontEndService) Monitor(ctx context.Context) error {
 		connectivityMap := map[string]bool{}
 		delay := 3 * time.Second // when started grant more time to write the whole config (minimize intial link flapping)
 		_ = fes.WaitStart(ctx)
+		var refreshCancel context.CancelFunc
 		for {
 			select {
 			case <-ctx.Done():
```
zolug, line 262: Missing refreshCancel() call.

pkg/ambassador/tap/stream/stream.go
```diff
@@ -224,3 +231,56 @@ func (s *Stream) open(ctx context.Context) error {
...
+		}
+		// found current target
+		// if target is enabled then everything is fine
+		if target.Status == nspAPI.Target_ENABLED {
+			return nil
+		}
+		break
+	}
+	// Target is disabled since the NSP has set its status to disable
+	// during refresh. This might be because the NSP considered the refresh
```
zolug, line 263: Could you pls elaborate this a bit more. I don't get the explanation.

pkg/ambassador/tap/stream/stream.go
```diff
@@ -224,3 +231,56 @@ func (s *Stream) open(ctx context.Context) error {
 	}
 	return nil
 }
+
+func (s *Stream) refresh(ctx context.Context) error {
+	err := s.TargetRegistry.Register(ctx, s.getTarget())
```
zolug, line 236: I don't get this function, some description would be appreciated about the motives here.
For me it seems the Target gets saved into the registry (overriding any existing entry), then the list of Targets are fetched from the registry to compare the IP list. Why would they differ if the target got just saved into the registry based on the stream?

LionelJouin: Yes, the NSP overwrite the status to `DISABLED` for newly registered targets, so the tapa has to verify if it is still in the `ENABLED` status, if it is, then it is fine and the refresh worked on time, but if it is not, it would mean the target has to register a new time (so it call the open function).
The Register function is a function that takes registration and updates, so NSP would consider it as a new registration (just it will replace the status since `ENABLED` is not allowed for new registration).

zolug (approved):

PR #206, opened 2022-04-22
Fix go routine leak with retry trigger

LionelJouin:
To cast a time channel to struct channel, a new go routine is created. A select with a new ctx has been added to cancel the retry trigger if the retry function is cancelled.

### DIFF
--- a/pkg/retry/config.go
+++ b/pkg/retry/config.go
@@ -23,7 +23,7 @@ import (
 
 // RetryTrigger is the function definition specifying when the Do
 // function should continue in an new attempt.
-type RetryTrigger func() <-chan struct{}
+type RetryTrigger func(context.Context) <-chan struct{}
 
 // RetryCondition is the function definition specifying if the Do
 // function should continue in an new attempt or not.
@@ -38,8 +38,8 @@ type Config struct {
 func newConfig() *Config {
 	return &Config{
 		context: context.Background(),
-		retryTriggerFunc: func() <-chan struct{} {
-			return delay(10 * time.Millisecond)
+		retryTriggerFunc: func(ctx context.Context) <-chan struct{} {
+			return delay(ctx, 10*time.Millisecond)
 		},
 		retryConditionFunc: defaultRetryCondition,
 	}
@@ -49,10 +49,13 @@ func defaultRetryCondition(err error) bool {
 	return err != nil
 }
 
-func delay(d time.Duration) <-chan struct{} {
+func delay(ctx context.Context, d time.Duration) <-chan struct{} {
 	channel := make(chan struct{}, 1)
 	go func() {
-		<-time.After(d)
+		select {
+		case <-ctx.Done():
+		case <-time.After(d):
+		}
 		channel <- struct{}{}
 	}()
 	return channel
--- a/pkg/retry/option.go
+++ b/pkg/retry/option.go
@@ -41,8 +41,8 @@ func WithMaxAttempts(attempts uint) Option {
 // WithDelay sets a retry trigger with a 10 milliseconds delay
 // WithDelay sets the retry trigger function.
 func WithDelay(delayTime time.Duration) Option {
-	return WithRetryTrigger(func() <-chan struct{} {
-		return delay(delayTime)
+	return WithRetryTrigger(func(ctx context.Context) <-chan struct{} {
+		return delay(ctx, delayTime)
 	})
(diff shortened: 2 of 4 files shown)


uablrek:
How can one check the problem before this PR, and that it's fixed after it is applied.

LionelJouin:
I discovered the issue while running the tests of https://github.com/Nordix/Meridio/pull/205, maybe I should try to add tests in this pkg

uablrek:
The correction seems ok (but I don't like the double-context approach).

But I can't say that it LGTM because it doesn't. IMO the entire "retry" package is over engineered and should basically be removed. It suffers from "generic-disease". While trying to satisfy all possible use-cases, even those never used, the package has become hard to understand and therefore hard to maintain and use.

For reference I searched for other implementations and it look like some inspiration came from https://github.com/avast/retry-go.

What is needed is a delay function that handles context cancellation. I should look like;
```go
func Delay(ctx context.Context, d time.Duration) error {
	t := time.NewTimer(d)
	select {
	case <-ctx.Done():
		t.Stop()
		return fmt.Errorf("Interrupted")
	case <-t.C:
	}
	return nil
}
```
A `Timer` object is used to allow cleanup with `Stop()`. With this function retries can be implemented with a `for` loop. Simple and understandable for all programmers.

I checked the code where "retry" is used and it seems like the only used use-case is to allow retries to be cancelled by a context. I re-wrote the use of "retry" in [stream/loadbalancer.go](https://github.com/Nordix/Meridio/blob/b5800a016244097d0fbc236557cdb8cecd932c3e/pkg/loadbalancer/stream/loadbalancer.go#L104-L124) using the `Delay()` function;
```go
	go func() { // todo
		logrus.Debugf("Stream '%v' Start watchTargets", lb.GetName())
		var err error
		for {
			err = lb.watchTargets(lb.ctx)
			if retry.Delay(lb.ctx, 500*time.Millisecond) != nil {
				break
			}
		}
		if err != nil {
			logrus.Errorf("Stream '%v' watch Targets err: %v", lb.GetName(), err)
		}
	}()

	go lb.processPendingTargets(lb.ctx)
	var err error
	for {
		err = lb.watchFlows(lb.ctx)
		if retry.Delay(lb.ctx, 500*time.Millisecond) != nil {
			break
		}
	}
	return err
```

I am unsure if the best action is to accept this PR as temporary band-aid to save the beta release or to propose a restructure. Guess the first(?).

uablrek:
The example above are both `WithErrorIngnored()` which is extra good to replace with a for-loop. For limited retries a (child) context should be used. The wanted function is usually not "we must try exactly 10 times", but rather "we should try every 0.5s for 1m" in which case the number of attempts is irrelevant and the timeout should be handled by a `context.WithTimeout`. Here is an example of a retry function;
```go
func Retry(
	ctx context.Context, interval time.Duration,
	f func(ctx context.Context) error) error {

	if _, ok := ctx.Deadline(); !ok {
		return fmt.Errorf("Infinite retries")
	}

	var err error
	for {
		if err = f(ctx); err == nil {
			break
		}
		if Delay(ctx, interval) != nil {
			return err
		}
	}
	return err
}
```
But IMO it's usually better and clearer to do a for-loop directly where it's used. A `Retry()` function doesn't really help.

If the function `f()` need more parameters a method should be used, example from my unit-tests;
```go
type attempt int
func (a *attempt) tfunc(ctx context.Context) error {
	if *a <= 0 {
		return nil
	}
	*a--
	return fmt.Errorf("Fail")
}

	// Timeout!
	a = 12
	ctx, cancel = context.WithTimeout(context.Background(), time.Second)
	if err := Retry(ctx, 100*time.Millisecond, a.tfunc); err == nil {
		t.Fatalf("Expected timeout")
	}
	cancel()
```

LionelJouin:
Not sure if there is any other way to solve the problem than to use double-ctx. It is hidden for the users of the pkg except if they want to implement a new RetryTrigger, and in that case they might want to know if the ctx has been canceled or not to avoid goroutine leaks.

I tried to make this pkg as generic as possible so it could cover future use cases. It might make it more complex to maintain, but I don't think it is complex to use. I implemented it to add retries a bit everywhere in a common way. Here are all the places the retry pkg should be used: https://github.com/Nordix/Meridio/issues/147.

I think with a common way to handle retries it is more readable, and I don't think we should add dependencies for small features like that one.

PR #207, opened 2022-04-26
Trouble shooting guide

uablrek:
Add a Meridio Troubleshooting Guide.

### DIFF
--- a/docs/trouble-shooting/README.md
+++ b/docs/trouble-shooting/README.md
@@ -1,0 +1,509 @@
+# Meridio - Troubleshooting Guide
+
+This guide focuses on `Meridio`, the
+[Meridio-operator](https://github.com/Nordix/Meridio-Operator) is only
+briefly touched.
+
+To be able to troubleshoot *anything* you must know how it works when
+everything is OK. So try to follow this guide, check the logs and use
+the tools in a healthy (test) system first.
+
+We will often follow the traffic-path. In each POD in the path traffic can
+be traced, for instance with `tcpdump`. Normal Linux tools like `ip`
+can be used to check interfaces, addresses and routes. The state of
+some programs, e.g. [bird](https://bird.network.cz/) and
+[nfqlb](https://github.com/Nordix/nfqueue-loadbalancer/), can also be
+checked.
+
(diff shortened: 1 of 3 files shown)


docs/trouble-shooting/README.md
```diff
@@ -0,0 +1,342 @@
...
+kubectl exec -n red meridio-proxy-96nvc -- ifconfig
+# Or
+kubectl exec -it -n red meridio-load-balancer-5785674d6-4hqsh -c load-balancer -- sh
+```
+
+We will give examples how to troubleshoot individual Meridio
+containers below.
+
+
+### Ephemeral containers
```
LionelJouin, line 151: This feature is not available before Kubernetes 1.23 right?

docs/trouble-shooting/README.md
```diff
@@ -0,0 +1,342 @@
...
+
+--- 172.16.1.8 ping statistics ---
+1 packets transmitted, 1 received, 0% packet loss, time 0ms
+rtt min/avg/max/mdev = 0.859/0.859/0.859/0.000 ms
+```
+
+Traffic to the target goes via a "proxy".
+
+
+## Proxy
```
LionelJouin, line 298: The source routes have to be verified in the proxy. it should contains 1 rule per VIP with a table assigned. And in each table there should be nexthops with the IPs of the load-balancers

docs/trouble-shooting/README.md
```diff
@@ -0,0 +1,342 @@
...
+(fwmark) used to route traffic to different target PODs;
+
+```
+# kubectl exec -n red meridio-load-balancer-5785674d6-6cqhm -c load-balancer -- ip rule
+0:      from all lookup local
+96:     from all fwmark 0x5a lookup 90
+97:     from all fwmark 0x43 lookup 67
+98:     from all fwmark 0x3a lookup 58
+99:     from all fwmark 0x30 lookup 48
+100:    from 10.0.0.1 lookup 4096
```
LionelJouin, line 273: maybe This rule could be explained:
`100: from 10.0.0.1 lookup 4096`

docs/trouble-shooting/README.md
uablrek: Will add those

PR #212, opened 2022-04-29
Fix proxy bridge IP error after restart

LionelJouin:
After the proxy restarts, it was trying to re-add existing IPs to the bridge but an error was raised by netlink since the IP was already existing.

### DIFF
--- a/pkg/kernel/interface.go
+++ b/pkg/kernel/interface.go
@@ -17,6 +17,8 @@ limitations under the License.
 package kernel
 
 import (
+	"errors"
+	"os"
 	"reflect"
 
 	"github.com/nordix/meridio/pkg/networking"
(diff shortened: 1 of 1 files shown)


pkg/kernel/interface.go
```diff
@@ -93,6 +93,15 @@ func (intf *Interface) AddLocalPrefix(prefix string) error {
 	if err != nil {
 		return err
 	}
+	addrList, err := netlink.AddrList(i, netlink.FAMILY_ALL)
```
uablrek, line 96: This is a race. Probably never happens, but the address can in theory be added after the check. IMO it would be better to check the err and ignore any "already exist" error.

zolug: I also think checking for "already exist" error would be more straight forward.

LionelJouin: The error is a common error (syscall.Errno=file exists), if I ignore it, it will ignore all errors

LionelJouin: Sorry, I am not sure what I tried yesterday, but it works with the fix you proposed works fine. I modified the commit

PR #217, opened 2022-05-05
Documentation

LionelJouin:


readme.md
```diff
@@ -1,58 +1,73 @@
...
+![Overview](docs/resources/High-Level-Overview.svg)
+
+## Getting Started
+
+* [High Level Overview](docs/overview.md)
+* [Architecture and Concepts](docs/readme.md#table-of-contents)
+* [Quick Installation / Demo](docs/demo/readme.md)
+* Deployment
+* Contributing
+* Frequently Asked Questions
```
uablrek, line 30: Create links for the non-link items

uablrek:
Why is the external network called [VPN](https://en.wikipedia.org/wiki/Virtual_private_network) everywhere??

I would say that in >90% of the cases the external network is not a VPN. Maybe a site or operator local network, but not a VPN.

LionelJouin:
@uablrek I replaced `VPN` by `external network` and I added the links

uablrek:
Another general comment, not specific for this documentation;

Why use "consume" instead of "use"?

I don't "consume" an interface, I *use* it.

Example, I can "consume" a sandwich in which case it disappears, but when riding my bike I don't "consume a bicycle".

uablrek:
Cleanup all self-links like;
```
* [some unimplemented item]()
```
Use something like;
```
* some unimplemented item (TODO)
```

uablrek:
After all wining and nitpicking I must say that this is a great improvement of the Meridio documentation. I liked the NSP documents most.

LionelJouin:
Fixed. Great, thank you for your review. I will try to add the remaining items (https://github.com/Nordix/Meridio/issues/221) as soon as possible (in future PR).

PR #223, opened 2022-05-11
proxy; align Conduit MTU

zolug:
Proxy as a center of TAPA-Proxy and Proxy-LB xconnects
changes Network Service Requests by adding MTU (retrieved
through environment variable).

This approach does not require additional privileges on the TAPA side.
(Also, on the Operator side similarly to Resource Management by implementing a dedicated annotation the MTU value could be set through Conduit CR.)

### DIFF
--- a/cmd/proxy/internal/config/config.go
+++ b/cmd/proxy/internal/config/config.go
@@ -19,6 +19,7 @@ package config
 import (
 	"errors"
 	"net/url"
+	"strings"
 	"time"
 )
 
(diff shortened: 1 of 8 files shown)


LionelJouin (approved):
Looks good, I opened a ticket to improve the error in case the user set a wrong MTU. 
https://github.com/networkservicemesh/sdk-kernel/issues/471

Wondering what would happen if a user upgrade the proxy with a new MTU. I don't think NSM support MTU updates.

Issue #225, opened 2022-05-13
Load-balancer resiliency


LionelJouin:
Currently the load-balancer container cannot restart, the old nftables rules, and ip rules/tables are still in the pod after a restart which could cause issues, and the connection with the proxies are not recreated properly. The load-balancer pod no longer has the network interfaces but the proxies still have it.

PR #228, opened 2022-05-24
Fix Load-balancer resiliency

LionelJouin:
If the LB restarts, the rules and routes will remain in pod. So, before adding a target rule and route, the previous one are now removed.

New chain element for the full mesh client tracking the connection and the associated NSE name. So, if the connection has a problem (e.g. NSE Crash), NSM will set the connection to the same NSE and won't try to find a new one.

TODO: remove NSE from the map when they are removed (real close). This could be done using the NSM registry with the find function to remove the no longer existing NSEs from the map.

Issue: #225


zolug:
The knowledge of fullmesh client on available load-balancer NSEs could be worth utilizing in the newly interoduced tracker chain.

PR #232, opened 2022-06-02
Release IP on connection close

LionelJouin:
## Description
Release function in the IPAM server implemented
UnsetIPContext releasing IPs in the proxy

## Issue link
/

## Checklist

- Purpose
    - [ ] Bug fix
    - [x] New functionality
    - [ ] Documentation
    - [ ] Refactoring
    - [ ] CI
- Test
    - [ ] Unit test
    - [ ] E2E Test
    - [x] Tested manually
- Introduce a breaking change
    - [x] No
    - [x] No


pkg/ipam/server.go
```diff
@@ -118,7 +118,29 @@ func (is *IpamServer) Allocate(ctx context.Context, child *ipamAPI.Child) (*ipam
...
+	if child.GetSubnet().GetConduit() == nil {
+		return &emptypb.Empty{}, nil
+	}
+	if child.GetSubnet().GetConduit().GetTrench() == nil {
+		return &emptypb.Empty{}, nil
+	}
+	if trench.GetName() != getTrenchName(child.GetSubnet().GetConduit().GetTrench().GetName(), child.GetSubnet().GetIpFamily()) {
+		return &emptypb.Empty{}, nil
+	}
+	conduit, err := trench.GetConduit(ctx, child.GetSubnet().GetConduit().GetName())
```
zolug, line 135: I noticed that deleting/removing a conduit is not yet implemented. But once it will be in place, there might be a race between the watcher removing the conduit from the trench in ipam and this Release call.
GetConduit returns no error if the entry is not found. So the rest of the code might work on a nil conduit. Probably would be wise to add a check for that as well and just return.

Btw I might have forgotten most of the implementation details, but would it be possible to delete all the node allocations when the watcher observes that a conduit was removed from the system? So would trench.RemoveConduit() cleanup everyting?

zolug (approved):

PR #235, opened 2022-06-03
Enable readOnlyRootFilesystem=true

zolug:
- /tmp is provided through an emptyDir volume to host unix listeners
  (e.g. probes, NSM listenOn socket)
- removed NSM dnscontext client chain element as it required write
  access e.g. to /etc/resolv.conf, /etc/coredns/Corefile
- BIRD: emptyDir volumes
  - to write routing config (under etc/bird/)
  - to create unix socket to communicate with birdc (changed path
    from default /run/bird.ctl to /run/bird/bird.ctl)
  - to write log file and its backup (2 x 20 kbyte by default
    under /var/log/)

NSM NSC is not changed. (It uses dnscontext client, but currently
there's no way to change resolv.conf path in the NSC. However,
failure to writing resolv.conf won't result in a fatal error, so
readOnlyRootFilesystem=true could be set if really needed.)

Remote VLAN NSE works with readOnlyRootFilesystem=true without problems.
(No guarantees for future releases though.)

Note: Set 'medium' of emptyDir volumes to 'Memory'. Thus get mounted as tmpfs.
Meaning they contribute to the containers' memory consumption. (Must be
considered when setting memory resource requests/limits.)

### DIFF
--- a/cmd/frontend/internal/bird/bird.go
+++ b/cmd/frontend/internal/bird/bird.go
@@ -30,9 +30,24 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+var regexWarn *regexp.Regexp = regexp.MustCompile(`Error|<ERROR>|<BUG>|<FATAL>|<WARNING>`)
+var regexInfo *regexp.Regexp = regexp.MustCompile(`<INFO>|BGP session|Connected|Received:|Started|Neighbor|Startup delayed`)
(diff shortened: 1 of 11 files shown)


zolug:
https://github.com/Nordix/Meridio/issues/15

LionelJouin (approved):

Issue #234, opened 2022-06-03
Some targets are not receiving traffic after scaling


LionelJouin:
**Describe the bug**
The issue is similar to this one: https://github.com/Nordix/Meridio/issues/55
After scaling the targets, for instance from 4 to 5, some targets are not receiving traffic for x seconds while they are correctly configured in nfqlb with correct IP rules and IP routes. It's random, most of the time it works correctly, sometimes only 1 is receiving traffic, sometimes 2.

I tried with ctraffic and mconnect, the result is the same.

**To Reproduce**
1. Scale out targets
2. Wait x (like 20) seconds 
3. Send traffic

**Expected behavior**
All targets should receive the traffic.

**Context**
 - Kernel: 5.15
 - Kubernetes: v1.23.4
 - Spire: v1.2.2
 - Network Service Mesh: v1.3.1
 - Meridio: v0.5.1

Running on Kind, I haven't tried any other environment, but it happened to the internal CI team.

**Logs**
```
time=2022-06-03T11:08:45+02:00 level=info msg=load-balancer-trench-a-666f6bc986-99kpp: <nil> - Shm: tshm-stream-a
  Fw: own=0
  Maglev: M=9973, N=100
   Lookup: 30 68 68 57 16 16 16 75 30 16 16 16 75 16 16 75 30 16 16 16 75 16 30 75 68...
   Active: 17(16) 31(30) 58(57) 69(68) 76(75)

time=2022-06-03T11:08:45+02:00 level=info msg=load-balancer-trench-a-666f6bc986-99kpp: <nil> - 0:	from all lookup local
93:	from all fwmark 0x11 lookup 17
94:	from all fwmark 0x3a lookup 58
95:	from all fwmark 0x1f lookup 31
98:	from all fwmark 0x45 lookup 69
99:	from all fwmark 0x4c lookup 76
100:	from 20.0.0.1 lookup 4096
100:	from 40.0.0.0/24 lookup 4096
32766:	from all lookup main
32767:	from all lookup default

time=2022-06-03T11:08:45+02:00 level=info msg=load-balancer-trench-a-666f6bc986-99kpp: <nil> - default via 169.254.100.150 dev ext-vlan.100 table 4096 proto bird metric 32 
default via 172.16.0.50 dev load-balan-861a table 17 
default via 172.16.0.36 dev load-balan-861a table 31 
default via 172.16.1.44 dev load-balan-5f4e table 58 
[...]

PR #237, opened 2022-06-13
Fix e2e tests

LionelJouin:
## Description
- Update e2e test dependency
- Fix e2e tests
  - Better e2e traffic generator structure
  - mconnect as default traffic generator
  - Target tests changed since the API has changed
  - NFQLB Verifier
- UDP traffic test + init configuration
  - Add IPv4 and IPv6 UDP traffic tests
  - Add generic script to be defined for each env
  - Add Init configuration
- VIP configuration test
  1. Add VIP
  2. Test traffic on the new VIP
  3. Remove the VIP

## Issue link
https://github.com/Nordix/Meridio/issues/234

## Checklist

- Purpose
    - [ ] Bug fix
    - [ ] New functionality
    - [ ] Documentation
    - [x] Refactoring
    - [ ] CI
- Test
    - [ ] Unit test
    - [x] E2E Test
    - [ ] Tested manually
- Introduce a breaking change
    - [x] No
    - [x] No


uablrek:
IMO this is too KinD oriented. You should consider a potential Meridio user saying something like;

> Ok, we have Spire and NSM setup on our OpenStack K8s cluster (using SDN), now we want to execute the Meridio e2e as a conformance test.

E.g. addresses would probably differ and other things as well. A step on the way is to use separate `values.yaml` files for the setup rather than multiple `--set` parameters. Then tell the user to copy the `values.yaml` files from this repo and adapt them. (I did that for `xcluster`). An own `values.yaml` can be specified with the `-f my-values.yaml` to `helm`.

And `"using :latest is the poor man's chaos-monkey"`. Don't do that by default. Edit your private `values.yaml` files instead. (as an example at the moment of writing the fe:latest image seem broken).

LionelJouin:
This is not well documented yet, but right, this is the objective to have. Currently you need to have this environment set up in you cluster: https://github.com/Nordix/Meridio/blob/e2e/docs/test.md#initial-deployment (not showing, but it should include Spire, NSM...). The tests could then be run on any Kubernetes, Spire, NSM and Meridio version, so we could detect if there is incompatibilities between versions.

For now, what you need to do to run it in another environment is to create a new test.sh and modify the parameters to run the e2e tests. 
https://github.com/Nordix/Meridio/blob/e2e/Makefile#L84
https://github.com/Nordix/Meridio/blob/e2e/test/e2e/e2e_suite_test.go#L68

Here is an example with the Meridio-Operator, the tests are exactly the same, the test.sh and parameters are different (since the streams, conduit ... names are different and the way to add/remove a VIP using the operator is different than with the helm chart).
https://github.com/Nordix/Meridio-Operator/blob/e2e/test/e2e/data/kind/test.sh
https://github.com/Nordix/Meridio-Operator/blob/e2e/Makefile#L110

I am not sure how you run traffic in XCluster, but in kind I use this command: `docker exec -i {trench}` followed by the traffic generator command (mconnect or ctraffic) which will be added by the tests. `{trench}` will be replaced by the trench name where the traffic will be sent. Maybe alias of commands could be suggested to support all environment, or I could simply add 1 traffic generator command per trench as parameter.

Otherwise, probably I have to expose the VIP + ports used in case e2e test users are using different ones in their environment: https://github.com/Nordix/Meridio/blob/e2e/test/e2e/e2e_suite_test.go#L59

uablrek:
If the same `values.yaml` file is used for trench-a and b, then the VIP's will be the same on vlan=100 and 101. Is that intentional?

LionelJouin:
yes, this is fine, the trenches are separated network extensions, so they can have the same VIP addresses, it's a valid configuration.

PR #239, opened 2022-06-15
TAPA Multi-Conduit fixes

LionelJouin:
- Watch and Set only VIPs used in the conduit. The user should not use the same VIP in 2 conduits that belongs to the same trench (due to policy rules that will be conflicting)
- Fix Interface name. Before the interface name was set only to "nsc", so if 2 interfaces was requested, then only 1 was created. The interfacename chain element is required, kernel.NewClient has a bug if 2 connections have too identical id + NS name


cmd/tapa/main.go
```diff
@@ -140,7 +140,7 @@ func main() {
 			vfiomech.MECHANISM:   chain.NewNetworkServiceClient(vfio.NewClient()),
 			kernelmech.MECHANISM: chain.NewNetworkServiceClient(kernel.NewClient()),
 		}),
-		interfacename.NewClient("nsc", &interfacename.RandomGenerator{}),
+		interfacename.NewClient("nsm-", &interfacename.CounterGenerator{}),
```
zolug, line 143: > Before, the interface name was set only to "nsc", so if 2 interfaces were requested, then only 1 was created. The interfacename chain element is required, kernel.NewClient has a bug if 2 connections have too identical id + NS name. Now interfaces are in the format: nsm-<number>

The reason for "bug" with kernel.NewClient is that besides the ns name it relies on connection.Id to generate the interface name. Yet, connection.Id in TAPA does not have "random" prefix at its start, unlike the fullmesh client in proxy (which prepends the id to be used with uuid.New().String() ).

I also remember, that this intefaceName chain component was way too immature. For example it was generating new ids even if it wasn't necessary. I slightly remember it leading to problems, but can't recall the details. Maybe it was mostly related to the proxy when testing lb failovers or such.

On the other hand, letting nsm name the interfaces is not bulletproof either, mostly because of the fixed size of the interface name. Meaning that in case of "longer" ns names only the first 4 characters of the random string could be used to generate the interface name.

LionelJouin: right, the NSM generator is using the connection id, and in this multi-conduit case, the nsm interfaces will be the same if the conduit names are too similar (since the beginning of the connection id will be the conduit name). I also got the request the names should be in this format: nsm-<number>.
I think I also fixed the generation of new interface ids even if it wasn't necessary in [pkg/nsm/interfacename/common.go](https://github.com/Nordix/Meridio/pull/239/files#diff-b10cf177849d9666f321d3d2106f98085378b351f277314eae456f38df4a4a3f)

zolug: I see, I will run some tests then to see how the name generation performs.

pkg/nsm/interfacename/generator.go
```diff
@@ -51,3 +52,30 @@ func (rg *RandomGenerator) Generate(prefix string, maxLength int) string {
...
+func (cg *CounterGenerator) Generate(prefix string, maxLength int) string {
+	cg.mu.Lock()
+	defer cg.mu.Unlock()
+	if cg.usedNames == nil {
+		cg.usedNames = make(map[int]struct{})
+	}
+	selected := 0
+	length := maxLength - len(prefix)
+	length *= 10
+	length -= 1
```
zolug, line 71: Imho, max allowed decimal counter value should be (10^length)-1

Also, just noticed that MAX_INTERFACE_NAME_LENGTH is set to 16, while that is only true if the terminating null is included. So either change MAX_INTERFACE_NAME_LENGTH to 15, or use (10^(length-1))-1 as max counter value.

cmd/tapa/main.go
LionelJouin: this should be fixed now

pkg/nsm/interfacename/generator.go
LionelJouin: Fixed, thank you

pkg/nsm/interfacename/server.go
```diff
@@ -44,14 +43,14 @@ func NewServer(prefix string, generator NameGenerator) networkservice.NetworkSer
...
-	ine.SetInterfaceName(request, metadata.IsClient(ine))
+	ine.SetInterfaceName(request)
 	return next.Server(ctx).Request(ctx, request)
 }
 
 // Close it does nothing except calling the next Close in the chain
 // A non-nil error if a next element in the chain returns a non-nil error
 // It implements NetworkServiceServer for the interfacename package
 func (ine *interfaceNameServer) Close(ctx context.Context, conn *networkservice.Connection) (*empty.Empty, error) {
+	ine.UnsetInterfaceName(conn)
```
zolug, line 54: I think this mostly fits the TAPA use case. Although if the connect did not succeed at all, and the tapa user "disconnects" the trench-conduit-stream and does not intend to connect it again (i.e. connects a different conduit instead), then the interface name won't get released.
In general probably release on failed request would be more "fault proof", than caching. But I leave it up to you to decide.

LionelJouin: I am not sure if releasing on a request fail could work since the request can be used to update a connection. It is fails on an update, then the name will be released will it is still in use.

zolug: Right. Then probably some states would need to be associated with the interface names indicating if they belonged to established connections or not. And names of established connections would not be allowed to get removed upon failed Requests.
But I don't think we should spend more time on this one.

zolug (approved):

PR #241, opened 2022-06-27
Fix TAPA blocked Close if VIPs are updated at the same time

LionelJouin:
## Description

* If a Close is called, and the TAPA has to disconnect the conduit while the VIP addresses are updated at the same time, then the TAPA was blocked due to the wait the Stop function of the conduit configuration handler. The SetVIP was waiting for the mutex to be unlocked while the Disconnect was waiting for the Configuration Handler to be finished (which was stucked due to the SetVIP).
* Refactoring on the way the VIP and streams are handled. If the process takes too much time, the intermediate values are ignored (the values received between the very last reception and the one currently in process).

## Issue link

/

## Checklist

- Purpose
    - [x] Bug fix
    - [ ] New functionality
    - [ ] Documentation
    - [ ] Refactoring
    - [ ] CI
- Test
    - [ ] Unit test
    - [ ] E2E Test
    - [x] Tested manually
- Introduce a breaking change
    - [x] No
    - [x] No

### DIFF
--- a/pkg/ambassador/tap/conduit/configuration.go
+++ b/pkg/ambassador/tap/conduit/configuration.go
@@ -28,6 +28,10 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+const (
+	channelBufferSize = 1
+)
+
 type Configuration interface {
 	Watch()
 	Stop()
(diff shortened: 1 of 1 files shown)


zolug (approved):
The dead-lock seems to be gone.

PR #244, opened 2022-06-28
Non root user

zolug:
## Checklist

- Purpose
    - [ ] Bug fix
    - [x] New functionality
    - [ ] Documentation
    - [ ] Refactoring
    - [ ] CI
- Test
    - [ ] Unit test
    - [ ] E2E Test
    - [x] Tested manually
- Introduce a breaking change
    Meridio images now contain a numeric user that a container will be started with by default. Use "runAsUser: 0" to keep running as root.
    - [] No
    Deployment templates are modified to provide the required securityContext (runAsNonRoot and capabilities)
    - [ ] No

### DIFF
--- a/build/base/Dockerfile
+++ b/build/base/Dockerfile
@@ -1,4 +1,8 @@
 FROM alpine:3.15
-RUN apk update && apk add iproute2 tcpdump iputils
+RUN apk update && apk add iproute2 tcpdump iputils net-tools \
+  && setcap 'cap_sys_ptrace,cap_dac_override+ep' /sbin/ss \
+  && setcap 'cap_sys_ptrace,cap_dac_override+ep' /bin/netstat \
+  && setcap 'cap_net_raw+ep' /bin/ping \
+  && setcap 'cap_net_raw+ep' /usr/bin/tcpdump
 ADD https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/v0.4.2/grpc_health_probe-linux-amd64 /bin/grpc_health_probe
 RUN chmod a+x /bin/grpc_health_probe
--- a/build/ctraffic/Dockerfile
+++ b/build/ctraffic/Dockerfile
@@ -1,3 +1,7 @@
+ARG USER=tapa-user
+ARG UID=10006
+ARG HOME=/home/${USER}
+
 FROM golang:1.18.1 as build
 
 ENV GO111MODULE=on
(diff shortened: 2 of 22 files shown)


LionelJouin (commented):
The ctraffic container is only a testing container so the changes were not really needed in the yaml nor the dockerfile. Is it possible these changes affect the way to test/debug with the ctraffic container? 

Could you modify the documentation for the TAPA integration in the application pod? (debug capabilities can be avoided in this documentation)
https://github.com/Nordix/Meridio/blob/master/docs/user-application.md

Otherwise I can see new capabilities in the proxy, load-balancer, nsp, ipam and fe containers, are they used for debugging or they are required?
If they are required, then please update the list in the documentation:
https://github.com/Nordix/Meridio/blob/master/docs/proxy.md#privileges
https://github.com/Nordix/Meridio/blob/master/docs/load-balancer.md#privileges
https://github.com/Nordix/Meridio/blob/master/docs/nsp.md#privileges
https://github.com/Nordix/Meridio/blob/master/docs/ipam.md#privileges
https://github.com/Nordix/Meridio/blob/master/docs/front-end.md#privileges

Otherwise, I tried the changes, it works fine

zolug:
> The ctraffic container is only a testing container so the changes were not really needed in the yaml nor the dockerfile. Is it possible these changes affect the way to test/debug with the ctraffic container?

Yeah, I know, but it's also the container using/interacting with the tapa sidecar. So I wanted to showcase, that basically no capabilities are required to use it (except for the ones required by the debug tools or maybe the application that would rely on secondary networking).
Initially, I wanted to introduce a dedicated container to interact with the ambassador socket, but I discarded that idea.

Otherwise, tools like mconnect, ctraffic, ping, tcpdump, netstat, ss, strace, netcat (unless wanted to bind to ports < 1024) should be working fine. On the other hand, the container could be still ordered to run as root through the "runAsUser:0" sercurityContext option, in case you think it would be safer because of the e2e tests for example.

> Could you modify the documentation for the TAPA integration in the application pod? (debug capabilities can be avoided in this documentation) https://github.com/Nordix/Meridio/blob/master/docs/user-application.md

Sure, I will do that.

> Otherwise I can see new capabilities in the proxy, load-balancer, nsp, ipam and fe containers, are they used for debugging or they are required? If they are required, then please update the list in the documentation: https://github.com/Nordix/Meridio/blob/master/docs/proxy.md#privileges https://github.com/Nordix/Meridio/blob/master/docs/load-balancer.md#privileges https://github.com/Nordix/Meridio/blob/master/docs/nsp.md#privileges https://github.com/Nordix/Meridio/blob/master/docs/ipam.md#privileges https://github.com/Nordix/Meridio/blob/master/docs/front-end.md#privileges

Some of them are required by the main binary/binaries. I will update the docs.

> Otherwise, I tried the changes, it works fine

Have you maybe tried running e2e tests with it? Just to see, that e.g. ctraffic container provides access to all the necessary tools needed.

LionelJouin:
Great thank you. No I haven't tried with the e2e tests, I only tried manually.

LionelJouin (approved):

LionelJouin (approved):

zolug:
https://github.com/Nordix/Meridio/issues/15

PR #258, opened 2022-07-28
NSP Target Registry chained server

LionelJouin:
## Description

* New target registry server pattern allowing future feature, e.g.:
   * Logger / Metrics / Trace
   * KeepAlive
   * NSP multi-replicas support
* Current chain: watch handler -> store (sqlite + keepalive)

## Issue link

/

## Checklist

- Purpose
    - [ ] Bug fix
    - [ ] New functionality
    - [ ] Documentation
    - [x] Refactoring
    - [ ] CI
- Test
    - [x] Unit test
    - [ ] E2E Test
    - [x] Tested manually
- Introduce a breaking change
    - [x] No
    - [x] No


cmd/nsp/main.go
```diff
@@ -141,3 +131,34 @@ func main() {
...
+	}
+	keepAliveRegistry, err := keepAliveRegistry.New(
+		keepAliveRegistry.WithRegistry(sqlr),
+		keepAliveRegistry.WithTimeout(config.EntryTimeout),
+	)
+	if err != nil {
+		return nil, errors.Wrap(err, "Unable create keepalive registry")
+	}
+	return nsp.NewServer(
+		nspRegistry.NewServer(keepAliveRegistry),
```
zolug, line 148: Separation of nspRegistry and watchhandler seems forced to me. They both rely on the same keepAliveRegistry.

zolug:
What does "NSP multi-replicas support" stand for?

pkg/nsp/next/next.go
```diff
@@ -0,0 +1,67 @@
...
+func (ntrsi *NextTargetRegistryServerImpl) Unregister(ctx context.Context, target *nspAPI.Target) (*empty.Empty, error) {
+	if ntrsi.next == nil {
+		return &empty.Empty{}, nil
+	}
+	return ntrsi.next.Unregister(ctx, target)
+}
+
+// Watch will call the Watch function of the next chain element
+// If the next element is nil, then nil will be returned.
+func (ntrsi *NextTargetRegistryServerImpl) Watch(t *nspAPI.Target, watcher nspAPI.TargetRegistry_WatchServer) error {
```
zolug, line 58: Probably this is similar to NSM's Monitor thingy, but it seems only one chain entry is allowed to implement a "real" Watch function as it needs to block _forever_. This is probably not an issue, but easy to make mistakes if the user is not aware :)

(Also, I read somewhere that grpc streams are not thread-safe. So, for example let's say the backend i.e the db was partitioned. Then letting loose multiple go routines wouldn't work properly without synchronization. And probably it would require a special tail chain entry that would wait for the watch routines to keep the stream open.)

LionelJouin: Right, only one chain element would be allowed to implement the real watch function.

For the second part, do you mean we should have 1 goroutine to handle all watcher?

LionelJouin:
> What does "NSP multi-replicas support" stand for?

It means we could run multiple NSP instances at the same time. For instance, we could have a leader election system, the non-leaders instances will then redirect their call towards the leader one. I think separating the server into chain elements will simplify the development of this feature.

pkg/nsp/next/next.go
zolug: I just meant that special care should be taken in that case, and the default chaining couldn't be used as is.

Issue #245, opened 2022-06-28
monitor gRPC connection state


zolug:
**Is your feature request related to a problem? Please describe.**
gRPC connections towards NSP or IPAM servers might break for example due to some network issue.

Also, a working connection towards NSP server often contributes to container READINESS, so it would be beneficial if we could indicate changes in NSP connection through that.

A feature like this could have helped a lot in identifying a problem in an IPv6-only cluster, where the IPv6 DNS server became eventually unreachable in one of the proxy containers, and thus connection towards the NSP server failed to get re-established.

**Describe the solution you'd like**
Investigate how gRPC connection state to a service could be monitored, to be used by NSP clients to reflect it in their READINESS state.

**Describe alternatives you've considered**
NA

**Additional context**

zolug:
1., Replace grpc.Dial with grpc.DialContext if applicable
2., Pass grpc.WithBlock() option to grpc dial when applicable to block until an initial connection is established
3., Consider using experimental API grpc.WaitForStateChange() to reflect state changes through readiness probes

PR #268, opened 2022-08-16
Monitor NSP and IPAM client connection state changes

zolug:
## Description
Allow monitoring gRPC connection state changes e.g for IPAM and NSP clients, and adjust probe state respectively through context.

This might help in detecting connectivity issues affecting NSP or IPAM services.

The monitoring is lightweight meaning it works solely on the gRPC connection state maintained locally by the grpc pkg, no network traffic is involved. (Some grpc API functions marked as experimental were used, which might change in the future.)

## Issue link
#245 

## Checklist

- Purpose
    - [ ] Bug fix
    - [x] New functionality
    - [ ] Documentation
    - [ ] Refactoring
    - [ ] CI
- Test
    - [ ] Unit test
    - [ ] E2E Test
    - [x] Tested manually
- Introduce a breaking change
    - [x] No
      Readiness probe update for IPAM.
    - [] No

### DIFF
--- a/pkg/health/connection/connection.go
+++ b/pkg/health/connection/connection.go
@@ -1,0 +1,82 @@
+/*
+Copyright (c) 2022 Nordix Foundation
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package connection
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/nordix/meridio/pkg/health"
+	"github.com/sirupsen/logrus"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/connectivity"
+)
+
+type gRPCConnectionStateMonitor struct {
+	healthService string
+	*grpc.ClientConn
+}
+
+// Monitor -
+// Monitors connection state changes and updates respective health service through context.
+// Currently only gRPC connection is supported, which relies on EXPERIMENTAL API.
+func Monitor(ctx context.Context, healthService string, cc interface{}) error {
+	switch cc := cc.(type) {
+	case *grpc.ClientConn:
+		m := gRPCConnectionStateMonitor{
+			healthService: healthService,
+			ClientConn:    cc,
+		}
+		go func() {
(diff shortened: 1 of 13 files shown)


pkg/health/connection/connection.go
```diff
@@ -0,0 +1,87 @@
...
+	case *grpc.ClientConn:
+		m := gRPCConnectionStateMonitor{
+			healthService: healthService,
+			ClientConn:    cc,
+		}
+		go func() {
+			defer logrus.Debugf("Connection monitor exit (%s)", m.healthService)
+			for {
+				s := m.GetState()
+				if s == connectivity.Ready {
```
LionelJouin, line 49:
```suggestion
			health.SetServingStatus(ctx, m.healthService, s == connectivity.Ready)
```

pkg/health/connection/connection.go
```diff
@@ -0,0 +1,87 @@
...
+				// Block until connection state changes.
+				if !m.WaitForStateChange(waitCtx, s) {
+					// context got timeout or canceled
+					select {
+					case <-ctx.Done():
+						// main context done
+						return
+					default:
+						// timeout; try re-connect
+						m.Connect()
```
LionelJouin, line 78: This is already done automatically by GRPC no? Also, not sure if it is the right place to have this, since this pkg only advertise the state of the connection and not manipulating the connection, what do you think?

LionelJouin (approved):
Great feature, I tried it, it works fine.

The only missing part is the NSP connection from the IPAM: https://github.com/Nordix/Meridio/blob/master/docs/overview.md#communication
The connection is dialed in /pkg/ipam/trench/watcher.go (maybe it has to be moved in cmd?)

pkg/health/connection/connection.go
zolug: Unfortunately gRPC won't automatically "re-connect" for example after the server has disappeared temporarily.

This was dominant when I tested IPAM failover and noticed that readiness probe of proxies remained NOT_SERVING, until I forced some communication towards IPAM (e.g. by starting new targets) or simply restarted the proxy.

So even though I agree that this code should normally only monitor things, the whole feature wouldn't work properly with RPC protocols like IPAM which can be silent for long time and where there's no need for a stream to be opened all time (like watchers in case of NSP).

LionelJouin: But for instance if I delete the NSP pod, the load-balancer will be disconnected from the nsp service and will re-connect to it after the new pod is ready even without this `Connect()` call. Is it the GRPC state which has a bug or?

zolug: In case of NSP it will reconnect, because the lb wants to maintain a watcher so basically a stream all the time, which involves sending things on the gRPC connection. And like I said, gRPC try to set up a transport connection if none available upon a "send request" from the user.
So with gRPC you will get a transport connection on your initial Dial and any subsequent sends if there's none for whatever reason.

LionelJouin (approved):

Issue #253, opened 2022-07-20
Logging improvement needed


giangtraan:
Log produced by 

- [ ] IPAM, 
- [ ] NSP and 
- [ ] TAPA need improvement to give more information for troubleshooting.

uablrek:
@giangtraan Three sub-programs are specially mentioned in this issue. I assume there have been problems that were hard to find in these sub-programs. To be able to _improve_ logging (not only make it structured) I would like to know what those problems were.

PR #275, opened 2022-08-31
Structured logging

uablrek:
## Description

Structured logging a'la K8s, [go-logr/logr](https://github.com/go-logr/logr)

## Issue link

#253

## Checklist

- Purpose
    - [ ] Bug fix
    - [x] New functionality
    - [ ] Documentation
    - [ ] Refactoring
    - [ ] CI
- Test
    - [x] Unit test
    - [ ] E2E Test
    - [x] Tested manually
- Introduce a breaking change
    - [x] No
    - [x] No


uablrek:
#### Shall we make the ADP format the only one?

My approach was that Meridio is an open-source project first and Ericsson second. But let's be realistic, Meridio will be used only by Ericsson for a long time. As such we can make the Ericsson format the only format and add different formats later if needed. The current implementation can be seen as a PoC that different formats *can* be supported.

uablrek:
#### Shall we have a global logger?

I have kind of gone all-in on the [go-logr/logr](https://github.com/go-logr/logr) concepts (except for `Fatal()`). E.g. the `WithValues()` function can be used to pass adapted loggers down the food-chain. But the individual parts of Meridio are fairly small (at least compared to K8s) so a more pragmatic approach may be acceptable. For instance;

```go
// New() is supposed to be called once and will also set the global Logger
// (multiple calls to New() are allowed but will over-write the global Logger)
logger := log.New("Meridio-ipam", config.LogLevel)
// The global logger can be used anywhere
log.Logger.Info("Printed by the global logger")
```

N.B. the intended way of using [go-logr/logr](https://github.com/go-logr/logr) is still possible, but it is not enforced.

uablrek:
The latest commit implements an NSM logger, used as (see load-balancer/main.go);
```
	// Work-around for hard-codes logrus dependency in NSM
	if config.LogLevel == "TRACE" {
		nsmlog.EnableTracing(true)
		logrus.SetLevel(logrus.TraceLevel)
	}
	logger.Info("NSM trace", "enabled", nsmlog.IsTracingEnabled())
	ctx = nsmlog.WithLog(ctx, log.NSMLogger(logger)) // allow NSM logs
```
but trace logs are not formatted so I guess NSM uses it's own logger for trace.

uablrek:
#### Some thoughts about "subsystems"

These should be used with care (if used at all). It does not help any one (developers or users) to just make them up. IMO the only two real subsytems that make sense are `bird` and `nfqlb`. If we catch log printouts from those and use a "logger" to re-emit them in our choosen format (as we should IMO), _then_ a "subsystem" entry shall be used.

uablrek:
I have removed all conversion to structured logging in Meridio code and made this PR a pure add-on.

I would like a fast review/approval to allow independent and parallel conversion to structured logging of Meridio code.

uablrek:
The usage of the log package is removed from this PR but an example output can be found in https://github.com/Nordix/Meridio/pull/275#issuecomment-1238355515.

pkg/log/logger_test.go
```diff
@@ -0,0 +1,102 @@
+/*
```
LionelJouin, line 1: Add Copyright

pkg/log/logger.go
```diff
@@ -0,0 +1,202 @@
...
+	//zc.EncoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder (almost works)
+	zc.EncoderConfig.EncodeTime = timeEncoder
+	zc.EncoderConfig.EncodeLevel = levelEncoder
+	zc.Encoding = "json"
+	z, err := zc.Build()
+	if err != nil {
+		panic(fmt.Sprintf("Can't create a zap logger (%v)?", err))
+	}
+	return zapr.NewLogger(z.With(
+		zap.String("version", "1.0.0"), zap.Namespace("extra_data")))
```
LionelJouin, line 135: Is this supposed to be the Meridio Version?

go.mod
```diff
@@ -5,6 +5,8 @@ go 1.19
 require (
 	github.com/antonfisher/nested-logrus-formatter v1.3.1
 	github.com/edwarnicke/grpcfd v1.1.2
+	github.com/go-logr/logr v1.2.3
+	github.com/go-logr/zapr v1.2.0
```
LionelJouin, line 9: There is a new release 1.2.3 https://github.com/go-logr/zapr/releases

go.mod
```diff
@@ -21,6 +23,7 @@ require (
 	github.com/stretchr/testify v1.8.0
 	github.com/vishvananda/netlink v1.2.1-beta.2.0.20220630165224-c591ada0fb2b
 	go.uber.org/goleak v1.1.12
+	go.uber.org/zap v1.19.1
```
LionelJouin, line 26: There is a new release 1.23.0 https://github.com/uber-go/zap/releases

docs/logging.md
```diff
@@ -0,0 +1,201 @@
...
+`V(n)` method;
+
+```go
+logger.Info("This is a normal info message")
+logger.V(1).Info("This is a debug message")
+logger.V(2).Info("This is a trace message")
+```
+
+There is no defined "trace" level in output so both trace and debug
+messages will be have severity "debug". The level filtering is still valid
```
LionelJouin, line 98:
```suggestion
messages will have severity "debug". The level filtering is still valid
```

docs/logging.md
```diff
@@ -0,0 +1,201 @@
...
+though, trace messages are suspressed unless TRACE level is set.
```
LionelJouin, line 99:
```suggestion
though, trace messages are suppressed unless TRACE level is set.
```

LionelJouin (approved):

pkg/log/logger.go
uablrek: No, it's the version of the ADP format. It is _required_ according to the ADP format spec.

uablrek:
```json
  "required": [
    "version",
    "timestamp",
    "severity",
    "service_id",
    "message"
  ]
```

go.mod
uablrek: I have just done `go mod tidy`. How goes go select the versions?

LionelJouin: you can use `go mod edit -require go.uber.org/zap@v1.23.0` (or change manually in the go.mod) then `go mod tidy`, but otherwise I have no idea how go selects the versions.

uablrek: I edited go.mod

PR #325, opened 2022-11-23
Add a Meridio adapted initContainer image

uablrek:
## Description

An initContainer is used to set sysctl's in the load-balancer and proxy PODs. It is currently a BusyBox image, not related to any other Meridio image. This PR adds an init image based on the Meridio base image (same as all other Meridio images).

The initContainer will get more responsibilities in the future, e.g. to setup a tunnel device in test environments that have no secondary networks, like public-cloud and "standard" test clusters. Therefore a `meridio-init.sh` is included for now, but more advanced programs can be added if necessary.

The first commit in this PR only builds and uses the new init image. The start commands (a bunch of sysctl -w) is not altered. However the `meridio-init.sh` is prepared to handle start without parameters.

```yaml
      initContainers:
        - name: init
          image: registry.nordix.org/cloud-native/meridio/init:latest
          securityContext:
            privileged: true
```
In such case the script setup depending on `hostname`.
```sh
if test -n "$1"; then
	cmd=$1
	shift
else
	cmd=$INIT_FUNCTION
fi
if ! test -n "$cmd"; then
	# Make a guess from the hostname
	if hostname | grep -qE 'load-balance|stateless-lb'; then
		cmd=lb
	elif hostname | grep -q 'proxy'; then
		cmd=proxy
	fi
fi
```
It is however better to explicitly specify the function in an environment variable.
```yaml
      initContainers:
        - name: init
          image: registry.nordix.org/cloud-native/meridio/init:latest
          env:
            - name: INIT_FUNCTION
              value: "proxy"
          securityContext:
            privileged: true
```

### Tunnel setup

As mentioned above the first use of the init image is to setup a tunnel (vxlan) to the load-balancer (instead of a real external interface). The configuration may look something like:
```yaml
      initContainers:
        - name: init
          image: registry.nordix.org/cloud-native/meridio/init:latest
          env:
            - name: INIT_FUNCTION
              value: "lb"
             - name: TUNNEL_PARAMS
              value: "10.244.2.10,100"
         securityContext:
            privileged: true
```
The params are the the POD address of the local test-generator (TG) and the vxlan id. The id will be handled in the same way as the vlan-tag for test compatibility reasons. So in this example the tunnel device will be called `ext-vlan.100`, same as for vlan setup.

## Issue link

N/A

## Checklist

- Purpose
    - [ ] Bug fix
    - [x] New functionality
    - [ ] Documentation
    - [ ] Refactoring
    - [ ] CI
- Test
    - [ ] Unit test
    - [x] E2E Test
    - [x] Tested manually
- Introduce a breaking change
    - [x] No
    - [ ] No

### DIFF
--- a/hack/meridio-init.sh
+++ b/hack/meridio-init.sh
@@ -1,0 +1,206 @@
+#! /bin/sh
+##
+## meridio-init.sh --
+##
+##   This script should run in an initContainer and prepare for
+##   various Meridio PODs.
+##
+##   Tunnel handling copied from;
+##   https://github.com/Nordix/k8s-service-tunnel
+##
+##   Config:
+##
+##   Parameter:   Environment var:  Description:
+##   --dev=       TUNNEL_DEV        Tunnel device name (vxlan0)
+##   --master=    TUNNEL_MASTER     Tunnel master device (eth0)
+##   --peer=      TUNNEL_PEER       Ip address of the remote side
+##   --id=        TUNNEL_ID         VNI for vxlan (333)
+##   --dport=     TUNNEL_DPORT      Port for the remote side (5533)
+##   --sport=     TUNNEL_SPORT      Port for the local side (5533)
+##   --ipv4=      TUNNEL_IPV4       IPv4 on the tunnel device, e.g. 10.30.30.1/24
+##   --ipv6=      TUNNEL_IPV6       IPv6 on the tunnel device, e.g. fd00:1::1/64
+##   --ipv4-only  __ipv4_only       Only set sysctls for ipv4
+##   --ipv6-only  __ipv6_only       Only set sysctls for ipv6
+##
+## Commands;
+##
+
+prg=$(basename $0)
+dir=$(dirname $0); dir=$(readlink -f $dir)
+tmp=/tmp/${prg}_$$
+
+die() {
+    echo "ERROR: $*" >&2
+    rm -rf $tmp
+    exit 1
+}
+cmd_help() {
+    grep '^##' $0 | cut -c3-
+    rm -rf $tmp
+    exit 0
+}
+
+log() {
+	echo "$prg: $*" >&2
+}
+
+# initvar <variable> [default]
+#   Initiate a variable. The __<variable> will be defined if not set,
+#   from $TUNNEL_<variable-upper-case> or from the passed default
+initvar() {
+	local n N v
+	n=$1
+	v=$(eval "echo \$__$n")
+	test -n "$v" && return 0	# Already set
+	N=$(echo $n | tr a-z A-Z)
+	v=$(eval "echo \$TUNNEL_$N")
+	if test -n "$v"; then
+		eval "__$n=$v"
+		return 0
+	fi
+	test -n "$2" && eval "__$n=$2"
+	return 0
+}
+
+##   env
+##     Print environment.
+cmd_env() {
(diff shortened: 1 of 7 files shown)


uablrek:
A better place than `hack/meridio-init.sh` is requested.

uablrek:
@LionelJouin Will the addition of an `init` image require CI updates?

LionelJouin:
Yes, I could update the CI.

uablrek:
The tunnel procedure should be unified with
https://github.com/Nordix/k8s-service-tunnel

hack/meridio-init.sh
```diff
@@ -0,0 +1,212 @@
...
+##
+if test -n "$1"; then
+	cmd=$1
+	shift
+else
+	cmd=$INIT_FUNCTION
+fi
+if ! test -n "$cmd"; then
+	# Make a guess from the hostname
+	if hostname | grep -qE 'load-balance|stateless-lb'; then
```
LionelJouin, line 178: I don't know what users will use as attractor and conduit name, so I would more consider undefined "cmd" as a miss-configuration.

LionelJouin (commented):
I tried it, it works fine.

hack/meridio-init.sh
uablrek: I have removed the hostname guessing. I wasn't happy with it myself.

uablrek:
The init-container has been updated to be more like https://github.com/Nordix/k8s-service-tunnel.

The `$TUNNEL_PARAMS` is removed and is replaced by more clear `TUNNEL_*` variables. The function is tested in `xcluster`, using [ovl/forwarder-test](https://github.com/Nordix/nsm-test/tree/master/ovl/forwarder-test) the external connectivity can be selected with `--exconnect=vlan|multus|tunnel` for the same tests. There were a few suprises along the way, and based on new experiences this is what has to be specified;

```yaml
      initContainers:
        - name: meridio-init
          image: registry.nordix.org/cloud-native/meridio/init:local
          securityContext:
            privileged: true
          env:
            - name: INIT_FUNCTION
              value: lb
            - name: TUNNEL_PEER
              value: "192.168.1.202"
            - name: TUNNEL_ID
              value: "101"
            - name: TUNNEL_DEV
              value: "nsm-1"
            - name: TUNNEL_IPV4
              value: "10.30.101.1/24"
            - name: TUNNEL_IPV6
              value: "fd00::10.30.101.1/120"
```
The TUNNEL_ID (VNI) and the addresses must be coordinated with the external machine that has the "other" tunnel endpoint. This means that an IPAM in the cluster will not help. The addresses can be about anything that not causes problems on the external machine.

The VNI must be specified for the same reasons. If the external machine has several tunnels (representing secondary networks) they must have different VNIs.

The TUNNEL_DEV is what the FE sees and must be the same as the NFE_EXTERNAL_INTERFACE.

uablrek:
The operator config *may* be extended with something like;
```yaml
  interface:
    name: tunnel0
    type: tunnel
    tunnel:
      peer: "192.168.1.202"
      id: "101"
      local-ipv4: "10.30.101.1/24"
      local-ipv6: "fd00::10.30.101.1/120"
```

uablrek:
Testing revealed that sometimes the external tunnel end must be "excercised" to trig some UDP traffic so the connection through the service is setup. A `ping` towards the TUNNEL_IPV4 on the cluster-end of the tunnel is sufficient.

Issue #316, opened 2022-11-01
E2E tests instability


LionelJouin:
**Describe the bug**
Tests are failing randomly 10 to 20% of the time.

**To Reproduce**
/

**Expected behavior**
/

**Logs**
```
[38;5;13m•! [PANICKED] [190.727 seconds][0m
Scaling
[38;5;243m/home/jenkins/nordix/slave_root/workspace/meridio-e2e-test-kind/71/test/e2e/scaling_test.go:30[0m
  With one trench containing a stream with 2 VIP addresses and 4 target pods running
  [38;5;243m/home/jenkins/nordix/slave_root/workspace/meridio-e2e-test-kind/71/test/e2e/scaling_test.go:32[0m
    when scaling targets up by 1
    [38;5;243m/home/jenkins/nordix/slave_root/workspace/meridio-e2e-test-kind/71/test/e2e/scaling_test.go:148[0m
      [38;5;13m[1m[It] should receive the traffic correctly[0m
      [38;5;243m/home/jenkins/nordix/slave_root/workspace/meridio-e2e-test-kind/71/test/e2e/scaling_test.go:152[0m

  [38;5;243mBegin Captured GinkgoWriter Output >>[0m
    [1mSTEP:[0m Checking if all targets have receive ipv4 traffic with no traffic interruption (no lost connection) [38;5;243m11/01/22 16:09:02.33[0m
    [1mSTEP:[0m Checking if all targets have receive ipv6 traffic with no traffic interruption (no lost connection) [38;5;243m11/01/22 16:09:03.53[0m
  [38;5;243m<< End Captured GinkgoWriter Output[0m

  [38;5;13mTest Panicked[0m
  [38;5;13mIn [1m[It][0m[38;5;13m at: [1m/home/jenkins/nordix/slave_root/workspace/meridio-e2e-test-kind/71/test/e2e/utils/trafficgenerator.go:96[0m

  [38;5;13minvalid character 'R' looking for beginning of value[0m

  [38;5;13mFull Stack Trace[0m
    github.com/nordix/meridio/test/e2e/utils.(*MConnect).AnalyzeTraffic(0x15fa73d?, {0xc0000fb000, 0x6c9, 0x1000})
    	/home/jenkins/nordix/slave_root/workspace/meridio-e2e-test-kind/71/test/e2e/utils/trafficgenerator.go:96 +0x1a6
[...]

PR #334, opened 2022-12-06
Clean target IP from Neighbor table on add

LionelJouin:
## Description

If a target is disconnecting and connecting again to a conduit on a short period, the IP(s) allocated to the target might be the same, but the MAC address will be different. The LBs will still have the old IP(s)-MAC address association in the neighbor table, it then has to be cleaned, traffic towards that target will be lost until the time for the cache to be automatically removed.

This fix solves most of the failing e2e caused by: https://github.com/Nordix/Meridio/issues/234
Based on the latest 130 runs, we have now 13% of failing runs (30% without the fix).

1. Currently the neighbor entry for a target is cleaned no matter if the target is already configured for another stream or not. This means some traffic can go through stream-a (target-1 has opened it), and a target-1 opens stream-b, then target-1 will still get its neighbor entry in the LBs. Will it impact traffic on stream-a? Do we have to fix it?

## Issue link

https://github.com/Nordix/Meridio/issues/316
https://github.com/Nordix/Meridio/issues/234

## Checklist

- Purpose
    - [x] Bug fix
    - [ ] New functionality
    - [ ] Documentation
    - [ ] Refactoring
    - [ ] CI
- Test
    - [ ] Unit test
    - [x] E2E Test
    - [ ] Tested manually
- Introduce a breaking change
    - [x] No
    - [x] No

### DIFF
--- a/pkg/kernel/fwmark.go
+++ b/pkg/kernel/fwmark.go
@@ -68,6 +68,9 @@ func (fwmr *FWMarkRoute) configure() error {
 		return err
 	}
 
+	// Old ARP/NDP entry for that IP address could cause temporary issue.
+	fwmr.cleanNeighbor()
+
 	fwmr.route = &netlink.Route{
 		Gw:    fwmr.ip.IP,
 		Table: fwmr.tableID,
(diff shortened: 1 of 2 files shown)


uablrek:
To clear the neighbor (arp) cache will not impact traffic, except a slight increased latency while the cache is updated.

LionelJouin:
Do you think I should verify if a route towards the target to add already exists somewhere in the pod and clean its neighbor entry only if there is no other route? Or should I keep it as it is in this PR?
(a route towards the target to add will exists if that target has other streams opened)

uablrek:
IMO, keep it simple. To clear the cache will always work, while any intelligent selection algorithm may fail (and is harder to maintain)

PR #343, opened 2023-01-09
New e2e tests

LionelJouin:
## Description

The configuration of the new tests has been done only for the operator environment. The tests for the Helm environment have been disabled. 

- E2E new max targets test and error handling improved
   - Error handler for the e2e has been improved to catch all errors
   - New e2e test for the max-targets field in the stream (Disabled for now due to TAPA bugs)
- Add second conduit as e2e test init config
   - Initial e2e configuration now requires 2 conduit in the same trench.
   - Add new TAPA test to open a stream in a second conduit
- Add new new-stream test
   - new-stream tests the stream creation/deletion
- Add second stream e2e test
   - Configuration e2e test renamed to Vip
   - New second stream e2e test for TAPA
- Flow e2e tests
   - new-flow test adding a flow
   - flow-priority
- nsm-vlan attractor e2e test  
   - new-attractor-nsm-vlan e2e test added.
- Fix e2e destination port vars and fix helm env
   - All destination port variables for the e2e tests are now renamed with simplier name
   - Helm chart is now running with only supported tests
- Dest port range and port nat e2e tests
   - Flow test of the destination port range
   - Conduit port nat test
- e2e flow byte-match test
   - Add e2e test for the byte-match flow property
- Fix e2e test checking nfqlb in LBs
   - It could happen error was checked when verifying nfqlb in the LBs while the stream was not even configured.

https://github.com/Nordix/Meridio/blob/e2e-extended/docs/test.md#list-of-tests

#### open-second-stream
```
STEP: Selecting the first target from the deployment with label app=target-a in namespace red @ 01/02/23 23:35:14.308
STEP: Opening stream stream-a-ii (conduit: conduit-a-1, trench: trench-a) in target target-a-7fd479c8f5-5kf52 in namespace red @ 01/02/23 23:35:14.311
STEP: Waiting the stream to be opened in pod target-a-7fd479c8f5-5kf52 using ./target-client watch @ 01/02/23 23:35:14.362
STEP: Waiting for nfqlb in the stateless-lb-frontend-attractor-a-1-7c86c8db64-bspbz (red) to have 1 targets configured @ 01/02/23 23:35:14.906
STEP: Waiting for nfqlb in the stateless-lb-frontend-attractor-a-1-7c86c8db64-jhzv7 (red) to have 1 targets configured @ 01/02/23 23:35:14.951
STEP: Sending tcp traffic from the TG trench-a (red) to 20.0.0.1:4000 @ 01/02/23 23:35:15.041
[...]

PR #346, opened 2023-01-13
Fix TAPA concurrency issue

LionelJouin:
## Description

- TAPA: Stream retry delay configurable
   - The delay was 50ms before, even in the unit tests. It is now 1 second by default and 1ms in the unit tests.
- TAPA: fix concurrency issue in the stream retry
   - if multiple events occur (AddStream, SetStreams), a stack of Open would pile up waiting for the mutex to be unlocked. When the mutex would be unlocked (on close call), the close could then be executed, and after, piled open calls would execute again, which would cause the stream to be re-opened.
- TAPA: fix conduit SetVips
   - SetVips could be called even after the conduit has been disconnected. So the NSM connection would have been requested a new time and the interface would be remaining in the pod.

## Issue link

https://github.com/Nordix/Meridio/issues/316
https://github.com/Nordix/Meridio/pull/343 (stream-max-targets test)

## Checklist

- Purpose
    - [x] Bug fix
    - [ ] New functionality
    - [ ] Documentation
    - [ ] Refactoring
    - [ ] CI
- Test
    - [x] Unit test
    - [x] E2E Test
    - [x] Tested manually
- Introduce a breaking change
    - [x] No
    - [x] No

### DIFF
--- a/pkg/ambassador/tap/conduit/conduit.go
+++ b/pkg/ambassador/tap/conduit/conduit.go
@@ -55,13 +55,13 @@ type Conduit struct {
 	Configuration        Configuration
 	StreamManager        StreamManager
 	NetUtils             networking.Utils
-	StreamFactory        StreamFactory
-	connection           *networkservice.Connection
-	mu                   sync.Mutex
-	localIPs             []string
-	ctx                  context.Context
-	cancel               context.CancelFunc
-	logger               logr.Logger
+	// RetryDelay corresponds to the time between each Request call attempt
+	RetryDelay    time.Duration
+	StreamFactory StreamFactory
+	connection    *networkservice.Connection
+	mu            sync.Mutex
+	localIPs      []string
+	logger        logr.Logger
 }
 
 // New is the constructor of Conduit.
@@ -83,12 +83,13 @@ func New(conduit *ambassadorAPI.Conduit,
 		NodeName:             nodeName,
 		NetworkServiceClient: networkServiceClient,
 		NetUtils:             netUtils,
+		RetryDelay:           1 * time.Second,
 		connection:           nil,
 		localIPs:             []string{},
 	}
 	c.StreamFactory = stream.NewFactory(targetRegistryClient, c)
 	c.StreamManager = NewStreamManager(configurationManagerClient, targetRegistryClient, streamRegistry, c.StreamFactory, PendingTime, nspEntryTimeout)
-	c.Configuration = newConfigurationImpl(c.SetVIPs, c.StreamManager.SetStreams, c.Conduit.ToNSP(), configurationManagerClient)
+	c.Configuration = newConfigurationImpl(c.StreamManager.SetStreams, c.Conduit.ToNSP(), configurationManagerClient)
 	c.logger = log.Logger.WithValues("class", "Conduit", "instance", conduit.Name)
 	return c, nil
 }
@@ -101,10 +102,9 @@ func (c *Conduit) Connect(ctx context.Context) error {
 	if c.isConnected() {
 		return nil
 	}
-	c.ctx, c.cancel = context.WithCancel(ctx)
 	c.logger.Info("Connect")
 	nsName := conduit.GetNetworkServiceNameWithProxy(c.Conduit.GetName(), c.Conduit.GetTrench().GetName(), c.Namespace)
-	connection, err := c.NetworkServiceClient.Request(c.ctx,
+	connection, err := c.NetworkServiceClient.Request(ctx,
 		&networkservice.NetworkServiceRequest{
 			Connection: &networkservice.Connection{
 				Id:             fmt.Sprintf("%s-%s-%d", c.TargetName, nsName, 0),
@@ -137,9 +137,6 @@ func (c *Conduit) Connect(ctx context.Context) error {
 // Disconnect closes the connection from NSM, closes all streams
 // and stop the VIP watcher
 func (c *Conduit) Disconnect(ctx context.Context) error {
-	if c.cancel != nil {
-		c.cancel()
-	}
 	c.logger.Info("Disconnect")
 	c.mu.Lock()
 	defer c.mu.Unlock()
@@ -192,7 +189,7 @@ func (c *Conduit) GetIPs() []string {
 }
 
 // SetVIPs checks the vips which has to be added or removed
-func (c *Conduit) SetVIPs(vips []string) error {
+func (c *Conduit) SetVIPs(ctx context.Context, vips []string) error {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 	if !c.isConnected() {
@@ -232,10 +229,9 @@ func (c *Conduit) SetVIPs(vips []string) error {
 		}
 		c.connection.Context.IpContext.Policies = append(c.connection.Context.IpContext.Policies, newPolicyRoute)
 	}
-	c.ctx, c.cancel = context.WithCancel(context.TODO())
 	// update the NSM connection
 	_ = retry.Do(func() error {
-		ctx, cancel := context.WithTimeout(c.ctx, 20*time.Second) // todo: configurable timeout
+		ctx, cancel := context.WithTimeout(ctx, 20*time.Second) // todo: configurable timeout
 		defer cancel()
 		request := &networkservice.NetworkServiceRequest{
 			Connection: &networkservice.Connection{
@@ -256,8 +252,8 @@ func (c *Conduit) SetVIPs(vips []string) error {
 		}
 		c.connection = connection
 		return nil
-	}, retry.WithContext(c.ctx),
-		retry.WithDelay(500*time.Millisecond))
+	}, retry.WithContext(ctx),
+		retry.WithDelay(c.RetryDelay))
 	c.logger.Info("VIPs updated", "vips", vips)
 	c.localIPs = getLocalIPs(c.connection.GetContext().GetIpContext().GetSrcIpAddrs(), vips)
 	return nil
--- a/pkg/ambassador/tap/conduit/manager.go
+++ b/pkg/ambassador/tap/conduit/manager.go
@@ -111,7 +111,7 @@ func (sm *streamManager) AddStream(strm *ambassadorAPI.Stream) error {
 		return nil
 	}
 	sr.setStatus(ambassadorAPI.StreamStatus_PENDING)
-	go sr.Open(nspStream)
+	sr.Open(nspStream)
 	return nil
 }
 
@@ -160,7 +160,7 @@ func (sm *streamManager) SetStreams(streams []*nspAPI.Stream) {
 	for _, sr := range sm.Streams {
 		nspStream := sm.getNSPStream(sr)
 		if nspStream != nil {
-			go sr.Open(nspStream)
+			sr.Open(nspStream)
 		} else {
 			ctx, cancel := context.WithTimeout(context.TODO(), sr.Timeout)
 			defer cancel()
@@ -186,7 +186,7 @@ func (sm *streamManager) Run() {
 	for _, sr := range sm.Streams {
 		nspStream := sm.getNSPStream(sr)
 		if nspStream != nil {
-			go sr.Open(nspStream)
+			sr.Open(nspStream)
 		} else {
 			sr.setStatus(ambassadorAPI.StreamStatus_UNDEFINED)
 		}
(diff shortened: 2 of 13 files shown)


LionelJouin:
I just found a similar problem with the conduit handling, I will add a fix as part of this PR

pkg/ambassador/tap/conduit/conduit.go
```diff
@@ -193,7 +204,10 @@ func (c *Conduit) GetIPs() []string {
 func (c *Conduit) SetVIPs(vips []string) error {
 	c.mu.Lock()
 	defer c.mu.Unlock()
-	if !c.isConnected() {
+	c.ctxMu.Lock()
```
uablrek, line 207: Should be a;
```go
if c.ctx != nil { }
```

uablrek: And why the double mutex'es in the first place? `c.mu` is already held, which ensures single thread.

pkg/ambassador/tap/conduit/conduit.go
```diff
@@ -137,9 +144,12 @@ func (c *Conduit) Connect(ctx context.Context) error {
 // Disconnect closes the connection from NSM, closes all streams
 // and stop the VIP watcher
 func (c *Conduit) Disconnect(ctx context.Context) error {
+	c.isDisconnecting.Store(true)
```
uablrek, line 147: Move the `c.mu.Lock()` a few lines up and drop the unnecessary extra ctxMu lock.

pkg/ambassador/tap/conduit/manager.go
```diff
@@ -238,43 +238,44 @@ type streamRetry struct {
...
 	// Set cancelOpen, so the close function could cancel this Open function.
 	sr.ctxMu.Lock()
+	if sr.cancelOpen != nil { // a previous open is still running
+		sr.ctxMu.Unlock()
+		return
+	}
 	var ctx context.Context
 	ctx, sr.cancelOpen = context.WithCancel(context.TODO())
 	sr.ctxMu.Unlock()
+	go func() {
```
uablrek, line 251: A golden rule for synchronization primitives is;

**Lock data, _never_ lock code!**

This locks code.

pkg/ambassador/tap/conduit/manager.go
```diff
@@ -285,9 +286,12 @@ func (sr *streamRetry) Close(ctx context.Context) error {
 	if sr.cancelOpen != nil {
 		sr.cancelOpen() // cancel open
 	}
-	sr.ctxMu.Unlock()
 	sr.mu.Lock()
-	defer sr.mu.Unlock()
+	defer func() {
```
uablrek, line 290: Recode without defer. And ask "are two mutex'es _really_ needed?"

uablrek:
In general the mutex strategy must be improved. Or rather, there must _be_ a mutex strategy. The code gives the impression that a number of "gotchas" has trigged a number of mutex hacks. Golden rule repeated:

**Lock data, _never_ lock code!**

LionelJouin:
All the comments seems to be arround the double mutex pattern, so I answer as a new comment:

The reason to have 2 mutexes is to be able to force disconnecting while an NSM request is in process (in SetVIPs / Connect functions).

When `Disconnect` will be called, it will set `isDisconnecting` to true, cancel the context (mutex required in case the cancel is set at the same time), and then wait, thanks to the other mutex for the other functions (`Connect`/`SetVIPs`) to be dome.

This pattern has been applied in the conduit handler (/pkg/ambassador/tap/conduit/conduit.go) and stream manager (/pkg/ambassador/tap/conduit/stream.go)

After discussions, this kind of problem must instead be fixed using channels and goroutine.

LionelJouin:
Here is a small struct with the logic removed but with the same behavior:
```go
package main

import (
	"context"
	"sync"
	"sync/atomic"
)

func main() {}

type MyType struct {
	ctx             context.Context
	cancel          context.CancelFunc
	IsConnected     bool
	mu              sync.Mutex // needed to lock IsConnected
	muCtx           sync.Mutex // needed to lock cancel
	isDisconnecting atomic.Bool
}

func (mt *MyType) Connect() {
	mt.mu.Lock()
	defer mt.mu.Unlock()

	mt.muCtx.Lock()
	mt.ctx, mt.cancel = context.WithCancel(context.Background())
	mt.muCtx.Unlock()

	mt.IsConnected = true
}

// Cannot be called if is not connected or if it is disconnecting
func (mt *MyType) Update() {
	mt.mu.Lock()
	defer mt.mu.Unlock()
	if !mt.IsConnected || mt.isDisconnecting.Load() {
		return
	}

	mt.muCtx.Lock()
	mt.ctx, mt.cancel = context.WithCancel(context.Background())
	mt.muCtx.Unlock()

	<-mt.ctx.Done()
	mt.IsConnected = true
}

// Will close the Connect/Update calls
func (mt *MyType) Delete() {
	mt.isDisconnecting.Store(true)
	mt.muCtx.Lock()
	mt.cancel()
	mt.muCtx.Unlock()

	mt.mu.Lock()
	defer mt.mu.Unlock()
	mt.ctx = nil
	mt.IsConnected = false
	mt.isDisconnecting.Store(false)
}
```

uablrek:
First, you don't have to lock a call to cancel [since there is a lock in the function alredy](https://cs.opensource.google/go/go/+/refs/tags/go1.20:src/context/context.go;drc=61f5a672edd6cce959a47e20dcfb59eaaa5ebd56;l=453), so remains the function pointer. But it's never cleared (set to nil), which it really should BTW, so its not a problem either.

Suggested Delete without muCtx
```go
func (mt *MyType) Delete() {
	mt.mu.Lock()
	defer mt.mu.Unlock()
	mt.isDisconnecting.Store(true)
	if mt.cancel != nil {
	  mt.cancel()
	}
	mt.ctx = nil
	mt.cancel = nil
	mt.IsConnected = false
	mt.isDisconnecting.Store(false)
}
```
I don't know if mt.isDisconnecting.Store should be with or without the lock held. In the original example it was set true _without_ the lock and false _with_ the lock. That may not be harmful, but it is inconsistent. (I have not compiled the code, so there may be bugs)

uablrek:
More flaws, or inconsistencies with the double-mutex'es;
```go
func (mt *MyType) Update() {
	mt.mu.Lock()
	defer mt.mu.Unlock()
	if !mt.IsConnected || mt.isDisconnecting.Load() {
		return
	}
// If Delete is called here the old ctx will be cancelled, but not the new one
	mt.muCtx.Lock()
	mt.ctx, mt.cancel = context.WithCancel(context.Background())
	mt.muCtx.Unlock()
// If Delete is called here the new ctx will be cancelled, but not the old one
	<-mt.ctx.Done() // Simulate infinite update

	mt.IsConnected = true
}
```

uablrek:
One more;
```go
func (mt *MyType) Connect() {
	mt.mu.Lock()
	defer mt.mu.Unlock()
// If Delete is called here you will have a SEGV
	mt.muCtx.Lock()
	mt.ctx, mt.cancel = context.WithCancel(context.Background())
	mt.muCtx.Unlock()

	mt.IsConnected = true
}
```

uablrek:
> After discussions, this kind of problem must instead be fixed using channels and goroutine.

Not for data locking! Then mutex'es _should_ be used.

But to block (and enqueue) multiple calls to a go-routing by using a mutex _should not_ be used in go. That is a different (worse) way, traditionally using integer-semaphores. In go a channel should be used

LionelJouin:
Thank you for your comments.

> First, you don't have to lock a call to cancel [since there is a lock in the function alredy](https://cs.opensource.google/go/go/+/refs/tags/go1.20:src/context/context.go;drc=61f5a672edd6cce959a47e20dcfb59eaaa5ebd56;l=453), so remains the function pointer. But it's never cleared (set to nil), which it really should BTW, so its not a problem either.

Yes, it should since the `Update` function reset the context at each called.

> I don't know if mt.isDisconnecting.Store should be with or without the lock held. In the original example it was set true without the lock and false with the lock. That may not be harmful, but it is inconsistent. (I have not compiled the code, so there may be bugs)

Right, it must be locked using muCtx so there is no problem with multiple Delete calls. 

> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx

Right, it that case, I checked, it can simply be removed from the struct since the ctx, is not used anywhere else. Only Cancel is important.

> Not for data locking! Then mutex'es should be used! But to block (and enqueue) multiple calls to a go-routine by using a mutex should not be used in go. That is a different (worse) way, traditionally using integer-semaphores. In go, a channel should be used

I still don't really see how to implement this.

This is how I rewrote it:

```
package main

import (
	"context"
	"sync"
)

func main() {}

type MyType struct {
	cancel          context.CancelFunc
	IsConnected     bool
	mu              sync.Mutex // needed to lock IsConnected and Connect/Update/Disconnect logic
	muCtx           sync.Mutex // needed to lock cancel and isDisconnecting
	isDisconnecting bool
}

func (mt *MyType) Connect() {
	mt.mu.Lock()
	defer mt.mu.Unlock()

	mt.muCtx.Lock()
	if mt.isDisconnecting {
		mt.muCtx.Unlock()
		return
	}
	if mt.cancel != nil {
		mt.cancel()
	}
	_, cancel := context.WithCancel(context.Background())
	defer cancel()
	mt.cancel = cancel
	mt.muCtx.Unlock()

	// Connect logic here

	mt.IsConnected = true
}

// Cannot be called if is not connected or if it is disconnecting
func (mt *MyType) Update() {
	mt.mu.Lock()
	defer mt.mu.Unlock()

	if !mt.IsConnected {
		return
	}

	mt.muCtx.Lock()
	if !mt.IsConnected || mt.isDisconnecting {
		mt.muCtx.Unlock()
		return
	}
	if mt.cancel != nil {
		mt.cancel()
	}
	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()
	mt.cancel = cancel
	mt.muCtx.Unlock()

	<-ctx.Done() // Update logic here

	mt.IsConnected = true
}

// Will close the Connect/Update calls
func (mt *MyType) Delete() {
	mt.muCtx.Lock()
	if mt.isDisconnecting {
		mt.muCtx.Unlock()
		return
	}
	mt.isDisconnecting = true
	if mt.cancel != nil {
		mt.cancel()
	}
	mt.muCtx.Unlock()

	mt.mu.Lock()
	defer mt.mu.Unlock()

	if !mt.IsConnected {
		return
	}

	// Disconnect logic here
	mt.IsConnected = false

	mt.muCtx.Lock()
	mt.isDisconnecting = false
	mt.muCtx.Unlock()
}
```

uablrek:
There are several bugs in the code above, mostly multiple calls to cancel().

> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx

The point here is to allow the _caller_ to decide about cancel's or timeouts. The called function should honor the passed ctx and cleanup+return if the context is canceled.

```go
package main

import (
	"context"
	"sync"
)

func main() {}

type connState int

const (
	idle connState = iota
	connected
	deleted
	canceled
	failed
)

type MyType struct {
	mu              sync.Mutex
	state           connState
}


// The context should be passed by the caller to allow cancel (or
// timeouts) in the way the caller wishes.
func (mt *MyType) Connect(ctx context.Context) {
	mt.mu.Lock()
	defer mt.mu.Unlock()

	if mt.state != idle {
		return					// Can only connect once
	}
	mt.state = connected

	// Connect logic here which honors the ctx

	// (the state can be connected (happy path) or canceled/failed on return)
}

func (mt *MyType) Update(ctx context.Context) {
	mt.mu.Lock()
	defer mt.mu.Unlock()

	if mt.state != connected {
		return					// Can only update in connected state
	}

	// What was the logic here? I write as I assume it works below.
	//<-mt.ctx.Done() 

	// The Update logic. Report outcome via a channel (not a context)
	// if it is in a go-routine. The caller may cancel the context at
	// any time.

	res := make(chan int)

	// Update logic here which honors the ctx

	select {
	case <-ctx.Done():
		mt.state = canceled
		return
	case rc := <-res:
		if rc != 0 {
			mt.state = failed
			// (return an error here?)
		}
	}

	// (the state can be connected (happy path) or canceled/failed on return)
}
// Will close the Connect/Update calls
func (mt *MyType) Delete(ctx context.Context) {
	mt.mu.Lock()
	defer mt.mu.Unlock()

	if mt.state != connected {
		return					// Can only delete in connected state
	}
	mt.state = deleted
	
	// Disconnect logic here which honors the ctx, and a select a'la Update()

	// (if the state is deleted, canceled or failed on return probably
	// doesn't matter for Delete())
}
```

The functions should probably return an error.

LionelJouin:
Ok, it's simpler, but it's also very different from the behavior I wanted to reach. My goal was to cancel the Connect or Update when Delete was called.

uablrek:
I see. But the muCtx will not work as intended! Example:
```go
func (mt *MyType) Connect() {
	mt.mu.Lock()
	defer mt.mu.Unlock()
// Delete() called here. Cancel will NOT be called
	mt.muCtx.Lock()
	mt.ctx, mt.cancel = context.WithCancel(context.Background())
	mt.muCtx.Unlock()
// Delete called here works as intended
	mt.IsConnected = true
}
```

uablrek:
Here Delete() calls cancel. The a ctx-child (and it's cancel()) is stored in the struct, but it's created _once_ in `NewMyType()`.

```go
package main

import (
	"context"
	"fmt"
	"sync"
)

func main() {}

type connState int

const (
	idle connState = iota
	connected
	deleted
	canceled
	failed
)

type MyType struct {
	mu     sync.Mutex
	state  connState
	ctx    context.Context		// Might contain a logger
	cancel context.CancelFunc
}

func NewMyType(ctx context.Context) *MyType {
	var mt MyType
	mt.ctx, mt.cancel = context.WithCancel(ctx)
	return &mt
}

func (mt *MyType) Connect() error {
	mt.mu.Lock()
	defer mt.mu.Unlock()

	if mt.state != idle {
		return fmt.Errorf("Already connected (well, in-use at least)")
	}
	mt.state = connected

	// Connect logic here which honors the mt.ctx

	return nil
}

func (mt *MyType) Update() error {
	mt.mu.Lock()
	defer mt.mu.Unlock()

	if mt.state != connected {
		return fmt.Errorf("Not connected")
	}

	// Update logic here which honors the mt.ctx

	return nil
}

// Delete MUST be called after Connect() to ensure cancel() 
func (mt *MyType) Delete(ctx context.Context) {

	// Cancel any on-going Connect or Update
	// NO SYNC IS NEEDED HERE! MULTIPLE CALLS WORKS! THE POINTER IS !=nil
	mt.cancel()

	mt.mu.Lock()
	defer mt.mu.Unlock()

	if mt.state == deleted {
		return // Already deleted
	}

	// Here the state can be
	//  idle - do nothing (shouldn't happen)
	//  connected - disconnect. Normal case
	//  canceled - a Connect or Update was canceled (might be difficult)
	//  failed - a Connect or Update has failed. Do nothing?

	// Disconnect logic here which doesn't care about the mt.ctx
	// (since it's canceled), but honors the passed ctx.

	mt.state = deleted
}
```

LionelJouin:
still, you delete (ny bad, I called it like that in my example, while it must be disconnect), but re-connect doesn't work because the ctx is set in the New function

uablrek:
IMO you should create a new object on re-connect. But if you _really_ want to re-use the object you may refresh the ctx,cancel in Delete() and set state=idle. (and rename Delete -> Disconnect)

uablrek:
I think it's best to clean-up everything necessary in Connect() or Update() when cancelled or failed. Then the Disconnect() can disconnect if state==connected, and do nothing for states cancelled or failed.

uablrek:
The stored ctx is a _child_ ctx. Meaning that cancel() on the parent also cancels the child.

LionelJouin:
I fixed it based on your comments, I had to refactor somethings so now it follows the solution described here: https://github.com/Nordix/Meridio/pull/346#issuecomment-1421449603

The VIP Watcher has been moved from the conduit handler to the conduit manager (conduitRetry).

pkg/ambassador/tap/trench/conduit.go
```diff
@@ -21,66 +21,65 @@ import (
...
-		cc.status = connected
-		break
+		return nil
+	}, retry.WithContext(ctx),
+		retry.WithDelay(cc.RetryDelay))
+
+	cc.ctxMu.Lock()
+	defer cc.ctxMu.Unlock()
+	if ctx.Err() == nil {
+		cc.configuration.Watch()
```
uablrek, line 72: Is disconnect() supposed to work here without block? It _will_ be blocked until cc.configuration.Watch() returns.

uablrek (approved):
Please check the comment about blocking disconnect(). If that's as intended I approve.

uablrek:
I have not reviewed the tests, and I do not intend to :smile:

pkg/ambassador/tap/trench/conduit.go
LionelJouin: This is fine, Watch is not blocking

PR #347, opened 2023-01-17
Use structured logging on TRACE level

uablrek:
## Description

In NSM v1.7.x the NSM logger can be configured to use structured logging. This PR ensures that structured logging is used even for NSM trace logging.

The NSM trace logging uses a format with indentations that is awkward when structured logging is used. A better way would be to post-process log entries whic is made simple by structured logging.

## Checklist

- Purpose
    - [x] Bug fix
    - [ ] New functionality
    - [ ] Documentation
    - [ ] Refactoring
    - [ ] CI
- Test
    - [ ] Unit test
    - [ ] E2E Test
    - [x] Tested manually
- Introduce a breaking change
    - [x] No
    - [x] No

### DIFF
--- a/cmd/tapa/main.go
+++ b/cmd/tapa/main.go
@@ -83,6 +83,21 @@ func main() {
 	}
 
 	logger := log.New("Meridio-tapa", config.LogLevel)
+	logger.Info("Configuration read", "config", config)
+	if err := config.IsValid(); err != nil {
+		log.Fatal(logger, "config.IsValid", "error", err)
+	}
+
+	if config.LogLevel == "TRACE" {
+		nsmlog.EnableTracing(true) // enable tracing in NSM
+		logrus.SetLevel(logrus.TraceLevel)
+	}
+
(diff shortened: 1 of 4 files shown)


cmd/tapa/main.go
```diff
@@ -83,6 +83,21 @@ func main() {
...
 
 	logger := log.New("Meridio-tapa", config.LogLevel)
+	logger.Info("Configuration read", "config", config)
+	if err := config.IsValid(); err != nil {
+		log.Fatal(logger, "config.IsValid", "error", err)
+	}
+
+	if config.LogLevel == "TRACE" {
+		nsmlog.EnableTracing(true) // enable tracing in NSM
+		logrus.SetLevel(logrus.TraceLevel)
```
LionelJouin, line 93: Do we still need logrus?

uablrek: Yes, unfortunately. NSM requires this since they are checking internally.  I had a comment on it before, which I should bring back.

LionelJouin (approved):

PR #348, opened 2023-01-19
Fix NSP Target Registry on stream conduit switch

LionelJouin:
## Description

When a stream was switched between 2 different conduits (switched or deleted/created in a new conduit), the NSP sqlite registry was taking the old conduit association instead of updating the conduit name of the association.

This case is covered by the e2e test: `new-stream` followed with `new-attractor-nsm-vlan` 

### To Reproduce

To reproduce the bug:
1. Create a stream (S-1) in a conduit (C-1)
2. Open the stream (S-1) in a target
3. Move the stream (S-1) in conduit (C-2) 
4. Open the stream (S-1) in a target

### Reason

Due to the missing `FullSaveAssociations` in gorm setting, and the previous registration of a Target in S-1.C-1, when the stream will be moved, the NSP will register new targets with S-1.C-2 without updating the nested conduit property C-1 to C-2.
This will cause the NSP to return an empty list of targets when watching S-1.C-2. The LBs will then not configure any target for that stream and the TAPA will then keep trying to register itself with a new identifier.

### Limitations

Due the way the data are stored, 2 objects (Stream, Conduit, Trench) with the same name would cause problems.

A better solution would be to fix it via a new way to represent the data in the database. So we would need a v2 of the sqlite registry. Then we would need a way to recover and convert the v1 to v2 (to handle the upgrade), but downgrade would not be supported since a v2 to v1 could not be added.

### Alternative to the limitations

Since Meridio v0.5.0 and the introduction of the keepalive mechanism in the NSP target registry (https://github.com/Nordix/Meridio/pull/205), we could not really care about the backward compatibility of the database, and just switch to v2 without any recover/convert mechanism. The TAPA updates itself every 30 seconds, so it will re-register itself if it sees it has been deleted. During an upgrade, this means there would be a traffic interruption.

## Issue link

https://github.com/Nordix/Meridio/pull/343

## Checklist

- Purpose
    - [x] Bug fix
    - [ ] New functionality
    - [ ] Documentation
    - [ ] Refactoring
    - [ ] CI
- Test
    - [x] Unit test
    - [x] E2E Test
    - [ ] Tested manually
- Introduce a breaking change
    - [x] No
    - [x] No

### DIFF
--- a/pkg/nsp/registry/sqlite/sqlite.go
+++ b/pkg/nsp/registry/sqlite/sqlite.go
@@ -35,9 +35,10 @@ type TargetRegistrySQLite struct {
 	watchers map[*common.RegistryWatcher]struct{}
 }
 
-func New(datastore string) (types.TargetRegistry, error) {
+func New(datastore string) (*TargetRegistrySQLite, error) {
 	db, err := gorm.Open(sqlite.Open(datastore), &gorm.Config{
-		Logger: logger.Default.LogMode(logger.Silent),
+		Logger:               logger.Default.LogMode(logger.Silent),
+		FullSaveAssociations: true,
 	})
 	if err != nil {
 		return nil, err
@@ -53,6 +54,15 @@ func New(datastore string) (types.TargetRegistry, error) {
 	return targetRegistrySQLite, nil
 }
 
+func (trsql *TargetRegistrySQLite) Close() error {
+	sqlDB, err := trsql.DB.DB()
+	if err != nil {
+		return err
+	}
+	sqlDB.Close()
+	return nil
+}
+
 func (trsql *TargetRegistrySQLite) Set(ctx context.Context, target *nspAPI.Target) error {
 	trsql.mu.Lock()
 	defer trsql.mu.Unlock()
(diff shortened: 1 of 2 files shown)


LionelJouin:
I just tried and managed to reproduced the issue you mentioned. 
I tried both PR (this current one + https://github.com/Nordix/Meridio/pull/346) with the branch https://github.com/Nordix/Meridio/tree/e2e-extended-tapa-nsp-merged, and I haven't managed to reproduce the issue, so it seems it was so concurrency problems.

zolug:
[trench-2-cond.txt](https://github.com/Nordix/Meridio/files/10663603/trench-2-cond.txt)

I used the config found in the attached trench-2-cond.txt file.
With 1 example POD opening stream-a1 conduit=load-balancer-a1.
After closing the stream, I removed it, and re-added to load-balancer-b1

zolug:
> [trench-2-cond.txt](https://github.com/Nordix/Meridio/files/10663603/trench-2-cond.txt)
> 
> I used the config found in the attached trench-2-cond.txt file. With 1 example POD opening stream-a1 conduit=load-balancer-a1. After closing the stream, I removed it, and re-added to load-balancer-b1

The problem seems to be related to setting the optional max-targets value for the stream. And has nothing to do with the changes.

zolug (approved):

PR #349, opened 2023-01-23
Allow internal target IP change without close/reopen

LionelJouin:
## Description

- Allow internal target IP change without close/reopen
   - When a trench/conduit/stream is deleted and recreated, the target that was connected will again request the connections with the old parameters (IPs, GWs, routes...). The proxy might get different subnets allocated and will forget the target IPs in use. So this will cause the targets to have wrong/colliding IP addresses.
   - The proxy is now detecting if a target is requesting the connection with invalid IP addresses and will fix them together with the other parameters (GWs, routes...).
   - In addition, the tapa can now handle the internal IP changes and will register itself with the new IP assigned.
- Fix NSP reconnect in the TAPA in case of NSP service delete/create  
   - If the NSP service is re-created, the TAPA will take around 15 minutes to re-connect to the NSP service. Having the time field set in the keepalive parameters for GRPC forces the DNS to be refreshed.
- Add delete/recreate trench e2e test
   - New e2e test deleting the complete trench-a and recreating it.

### Other alternative

The TAPA stops to update the connections to request new VIPs and instead, the proxy does it. The proxy would watch the VIP addresses in the conduit it serves, and would update the connection with the latest VIP list. This feature has been added to NSM, but I never tried it, and I am not sure how, from the NSC side, we can be aware of the connection changes.

#### delete-create-trench
```
STEP: Deleting the trench trench-a @ 01/24/23 11:14:14.408
STEP: Recreate the trench trench-a @ 01/24/23 11:14:48.716
STEP: Sending tcp traffic from the TG trench-a (red) to 20.0.0.1:4000 @ 01/24/23 11:15:31.173
STEP: Sending tcp traffic from the TG trench-a (red) to [2000::1]:4000 @ 01/24/23 11:16:05.862
```

## Issue link

/

## Checklist

- Purpose
    - [x] Bug fix
    - [ ] New functionality
    - [ ] Documentation
    - [ ] Refactoring
    - [ ] CI
- Test
    - [x] Unit test
    - [ ] E2E Test
    - [x] Tested manually
- Introduce a breaking change
    - [x] No
    - [x] No

### DIFF
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -230,19 +230,23 @@ func removeOldIPs(ips []string, gateways []string) []string {
 		}
 		gws = append(gws, n)
 	}
-	ipsMap := listToMap(ips)
-	for ip := range ipsMap {
+	res := []string{}
+	for _, ip := range ips {
 		i, _, err := net.ParseCIDR(ip)
 		if err != nil {
 			continue
 		}
+		ipInGatewaySubnet := false
 		for _, net := range gws {
 			if net.Contains(i) {
-				delete(ipsMap, ip)
+				ipInGatewaySubnet = true
 			}
 		}
+		if !ipInGatewaySubnet {
+			res = append(res, ip)
+		}
 	}
-	return mapToList(ipsMap)
+	return res
 }
 
 // Tells if a contains all b items
@@ -262,15 +266,6 @@ func listToMap(l []string) map[string]struct{} {
 	res := map[string]struct{}{}
 	for _, s := range l {
 		res[s] = struct{}{}
-	}
-	return res
-}
-
-// convert the keys of a map to a list
-func mapToList(m map[string]struct{}) []string {
-	res := []string{}
-	for k := range m {
-		res = append(res, k)
 	}
 	return res
 }
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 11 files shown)


pkg/proxy/proxy.go
```diff
@@ -155,32 +150,138 @@ func (p *Proxy) SetIPContext(conn *networkservice.Connection, interfaceType netw
...
+func removeOldIPs(ips []string, gateways []string) []string {
+	gws := []*net.IPNet{}
+	for _, ip := range gateways {
+		_, n, err := net.ParseCIDR(ip)
+		if err != nil {
+			continue
+		}
+		gws = append(gws, n)
+	}
+	ipsMap := listToMap(ips)
```
zolug, line 233: What's the point of this conversion considering the rest of the function?

pkg/proxy/proxy.go
```diff
@@ -155,32 +150,138 @@ func (p *Proxy) SetIPContext(conn *networkservice.Connection, interfaceType netw
...
+		ipContext.SrcIpAddrs = srcIPAddrs
+		ipContext.DstIpAddrs = dstIpAddrs
+		ipContext.ExtraPrefixes = p.Bridge.GetLocalPrefixes()
+		return
+	}
+	// The request is an update
+	if contains(ipContext.ExtraPrefixes, p.Bridge.GetLocalPrefixes()) &&
+		contains(ipContext.SrcIpAddrs, srcIPAddrs) &&
+		contains(ipContext.DstIpAddrs, dstIpAddrs) {
+		return
```
zolug, line 180: only matches if there are no VIPs, right?

pkg/proxy/proxy.go
LionelJouin: true, I will solve it

pkg/proxy/proxy.go
LionelJouin: no, it tells if all seconds list is in first list, so VIP will be ignored.

zolug:
Seems to work fine. Tested with a delayed proxy NSE and new VIP set as well, so that call to setVIPs override the request before the connection could have been re-established by nsm heal.

zolug (approved):

Issue #358, opened 2023-02-06
Structured logging hides contents of interface objects


uablrek:
**Describe the bug**

Some (quite many) go objects in Meridio are really interfaces. When used for output in structured logging the contents is empty `{}`. Example;

https://github.com/Nordix/Meridio/blob/c19b1d2e8c5dec12639ee0570773724bc3859cea/pkg/loadbalancer/stream/loadbalancer.go#L190

**To Reproduce**

Check the load-balancer logs:

```
# kubectl logs -n red -c stateless-lb meridio-load-balancer-7549d49c9f-f28xz | jq
...
{
  "severity": "info",
  "timestamp": "2023-02-06T09:29:37.965+00:00",
  "service_id": "Meridio-LB",
  "message": "AddTarget",
  "version": "1.0.0",
  "extra_data": {
    "class": "LoadBalancer",
    "instance": "stream1",
    "target": {}
  }
}
```

**Expected behavior**

Useful logging

uablrek:
Interface classes *may* be needed, but they should not be over-used since they basically trashes structured logging.

An overhaul is needed to log specific getters in interface objects rather than the objects themselves. This goes somewhat against the idea of structured logging.

PR #360, opened 2023-02-07
Log contents of interface objects

uablrek:
## Description

Log contents of interface objects instead of the object itself in structured logging.

After a bit of fumbling I realized the right way is to define a custom `MarshalJSON` function for the _real_ class. Example:
```go
func (t *Target) MarshalJSON() ([]byte, error) {
	ts := struct {
		Identifier int      `json:"identifier"`
		IPs        []string `json:"ips"`
	}{
		t.getIdentifier(),
		t.nspTarget.GetIps(),
	}
	return json.Marshal(&ts)
}
```

## Issue link

#358

## Checklist

- Purpose
    - [x] Bug fix
    - [ ] New functionality
    - [ ] Documentation
    - [ ] Refactoring
    - [ ] CI
- Test
    - [ ] Unit test
    - [ ] E2E Test
    - [ ] Tested manually
- Introduce a breaking change
    - [x] No
    - [x] No

### DIFF
--- a/pkg/loadbalancer/stream/loadbalancer.go
+++ b/pkg/loadbalancer/stream/loadbalancer.go
@@ -433,7 +433,7 @@ func (lb *LoadBalancer) verifyTargets() {
 		}
 		err := lb.RemoveTarget(target.GetIdentifier())
 		if err != nil {
-			lb.logger.Error(err, "deleting target", "target", target.GetIdentifier())
+			lb.logger.Error(err, "deleting target", "target", target)
 		}
 		lb.addPendingTarget(target)
 	}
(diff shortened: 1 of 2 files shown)


uablrek:
More to come. I will squash commits before "Ready for review". It wasn't nearly as bad as I feared.

LionelJouin (approved):

uablrek:
It turns out that this isn't the way after all.

All implementations of an interface that wants to use structured logging should implement the [logr.Marshaler](https://pkg.go.dev/github.com/go-logr/logr#Marshaler) interface. It should not be part of the interface itself, but the real implementation types should define a `MarshalLog()` function. Example (not tested):

```go
func (t *Target) MarshalLog() any {
	ts := struct {
		Identifier int      `json:"identifier"`
		IPs        []string `json:"ips"`
	}{
		t.getIdentifier(),
		t.nspTarget.GetIps(),
	}
	return &ts
}
```

Issue #356, opened 2023-02-03
LB not re-configuring routes after target IP update


LionelJouin:
**Describe the bug**
If a target got its IP addresses updated, but keep registering to the NSP with the same identifier, then the LBs won't reconfigure the IP addresses (routes toward that target) and will keep the old ones.

**To Reproduce**
Happens sometimes with the `delete-create-trench` test that have been introduced as part of this PR https://github.com/Nordix/Meridio/pull/349
I observed the problem one time with this seed: 274598459 (I haven't retried it)

**Expected behavior**
The LBs should reconfigure the IP addresses (routes) for the targets.

**Context**
 - Kubernetes: v1.26.0
 - Network Service Mesh: v1.6.0
 - Meridio: latest

**Logs**
[logs.zip](https://github.com/Nordix/Meridio/files/10578139/logs.zip)

PR #363, opened 2023-02-09
Fix target changes on LB

LionelJouin:
## Description

If a target with an identifier was previously registered with IPs and is now changing its IPs (IP refresh, different target...), the LB can now handle that case by removing it completly and adding it again.

the internal setTargets is now removing old targets before adding new ones. It also checks if IPs for an identifier have changed or not.

## Issue link

https://github.com/Nordix/Meridio/issues/356

## Checklist

- Purpose
    - [x] Bug fix
    - [ ] New functionality
    - [ ] Documentation
    - [ ] Refactoring
    - [ ] CI
- Test
    - [ ] Unit test
    - [x] E2E Test
    - [ ] Tested manually
- Introduce a breaking change
    - [x] No
    - [x] No


pkg/loadbalancer/stream/loadbalancer.go
```diff
@@ -340,31 +341,39 @@ func (lb *LoadBalancer) setTargets(targets []*nspAPI.Target) error {
...
+		if !exists {
+			err := lb.RemoveTarget(identifier)
 			if err != nil {
 				errFinal = fmt.Errorf("%w; %v", errFinal, err)
 			}
+			continue
 		}
-	}
-	for identifier := range toRemoveTargetsMap { // targets to remove
+		if slice.ContainsSameStrings(target.GetIps(), newTarget.GetIps()) { // have the same IPs?
```
uablrek, line 361: This is a set-operation. The right way is to add a set package, not a slice package. Here are generic sets as implemented in K8s

https://github.com/kubernetes/apimachinery/blob/v0.26.1/pkg/util/sets/set.go#L24

target.GetIps() should really return a set (set[string]). This would make the code more maintainable. This may be a large update, but to implement (steal) a set package may be a good first step.

uablrek: Godoc https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#Set

uablrek: For ref https://github.com/golang/go/issues/57433

uablrek:
Please see https://github.com/Nordix/Meridio/pull/363#discussion_r1104768969. Sorry for the fuss, but I think it's better to replace slices that are sets with sets in the future and use set-operations, but for now since we use slices maybe we should stick with them. But slice functions should be generic when it's possible.

Issue #355, opened 2023-02-01
NSP watch notifications do not reach LB


zolug:
**Describe the bug**
The LB might end up in a situation where there's a seemingly working gRCP stream (and underlying TCP connection) to the NSP server for watching external connectivity related changes, yet no message is received on it from the NSP.

According to tcpdump messages are received by stateless-lb from NSP when an update/refresh is due for the FE targets.
However, after some time other than the keepalive related packets no traffic is visible when running tcpdump on the NSP side of the connection when provoking change in external connectivity.

Debugging a bit on the NSP side showed that the watcher client is known to the server, and the server triggers a send on it for the events subscribed.

Restarting the NSP does not seem to help, a new TCP connection will be established but the problem remains (i.e. no incoming notifications). Restarting the LB container seems to solve the problem though.

**To Reproduce**
1. I've only seen this issues after suspending (for a couple hours) and waking my laptop running a Meridio deployment in Kind.
2. After step 1, the LB containers of the LB-FE PODs get stuck in non-ready, even though the external connectivity recovers and gets reported to NSP. The LB logs normally show an external connectivity unavailable event with a timestamp when the cluster got resumed.

Same problem was observed recently in a "real" environment.

**Expected behavior**
NSP watcher in LB should continue getting notifications from NSP server.

**Context**
 - Kubernetes: v1.22.9
 - Meridio: 1.0.0 (older releases also affected)
 ...

**Logs**

zolug:
Also, observed on clusters:

1., Between FE and NSP: The cluster was left running for a couple days, and in the background the vpp-forwarders were occasionally redeployed.

FE failed to prolong its target registration with NSP to announce its availability.
There were no apparent errors in FE, although the retry package wrapping announcement to periodically renew the registration does not log any errors if run in a way to ignore them.
NSP logs showed that all of a sudden the FE still maintaining external connectivity has unregistered. According to tcpdump except for the tcp keepalives there was no FE originated exchange, although it should have occurred every 30 seconds (like in a normal case).
Deleting the NSP POD failed to resolve the issue. Killing the affected frontend process in the container however worked.

2., After restarting the the node running NSP:

[...]

PR #417, opened 2023-05-17
Frontend: Cleanup Monitor() function

uablrek:
## Description

There seems to be a bug in here somewhere. But I couldn't follow all logic so I made a general cleanup.

One thing that might have been a bug is the "break" in
https://github.com/Nordix/Meridio/blob/60a45b5a8254ada9843466871abccc62eb9f1f30/cmd/frontend/internal/frontend/service.go#L314

The Monitor can exit more or less silently and announceFe will not be called in intervals ever after.

## Issue link

May fix the NSP-FE issue described in https://github.com/Nordix/Meridio/issues/355#issuecomment-1514477574

## Checklist

- Purpose
    - [x] Bug fix
    - [ ] New functionality
    - [ ] Documentation
    - [x] Refactoring
    - [ ] CI
- Test
    - [ ] Unit test
    - [ ] E2E Test
    - [x] Tested manually
- Introduce a breaking change
    - [x] No

### DIFF
--- a/cmd/frontend/internal/frontend/service.go
+++ b/cmd/frontend/internal/frontend/service.go
@@ -84,6 +84,7 @@ func NewFrontEndService(ctx context.Context, c *feConfig.Config) *FrontEndServic
 		secretManager:        watcher.NewObjectMonitorManager(ctx, c.Namespace, sdb, secret.CreateSecretInterface),
 		authCh:               authCh,
 		logger:               logger,
+		config:               c,
 	}
 
 	if len(frontEndService.vrrps) > 0 {
@@ -129,6 +130,7 @@ type FrontEndService struct {
 	secretManager        watcher.ObjectMonitorManagerInterface // keeps track changes of Secrets
 	authCh               chan struct{}                         // used by secretDatabase to signal updates to FE Service
 	logger               logr.Logger
+	config               *feConfig.Config
 }
 
 // CleanUp -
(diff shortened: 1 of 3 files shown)


uablrek:
I added timeouts of 5s (hard-coded) for communication with the NSP

cmd/frontend/internal/frontend/service.go
```diff
@@ -288,95 +293,79 @@ func (fes *FrontEndService) Monitor(ctx context.Context, errCh chan<- error) {
...
-						}()
-						fes.announceVIP(ctx, errCh)
+				if !hasConnectivity {
+					hasConnectivity = true
+					health.SetServingStatus(ctx, health.EgressSvc, true)
+					logger.Info("Announce frontend")
+					if err := announceFrontend(ctx, fes.targetRegistryClient); err != nil {
+						logger.Error(err, "Announce frontend connectivity")
 					}
+					// refresh NSP entry (even if announceFrontend() above fails?)
```
zolug, line 336: Yes, even if the send towards NSP fails, the FE still has external connectivity. The FE should keep trying to inform NSP about it.

cmd/frontend/internal/frontend/service.go
```diff
@@ -288,95 +293,79 @@ func (fes *FrontEndService) Monitor(ctx context.Context, errCh chan<- error) {
...
+
+			// determine status of gateways and connectivity
+			status := fes.parseStatusOutput(protocolOut, bfdOut)
+
+			// Gateway availibility notifications
+			if status.NoConnectivity() {
+				// although configured at least one IP family has no connectivity
+				if hasConnectivity {
+					denounce = true
+					// should this be moved to "if denounce" above?
```
zolug, line 324: It could be moved there, yes. It reflects the status of external connectivity and is used by probes.

cmd/frontend/internal/frontend/service.go
```diff
@@ -258,25 +258,30 @@ func (fes *FrontEndService) Monitor(ctx context.Context, errCh chan<- error) {
...
-	//linkCh := make(chan string, 1)
 	go func() {
-		var once sync.Once
-		//defer close(linkCh)
 		// status of external connectivity; requires 1 GW per IP family if configured to be reachable
-		init, noConnectivity := true, true
+		hasConnectivity := false
 		connectivityMap := map[string]bool{}
 		delay := 3 * time.Second // when started grant more time to write the whole config (minimize intial link flapping)
 		_ = fes.WaitStart(ctx)
```
zolug, line 266: The following code from the start of Monitor() should be moved after WaitStart() as well. The error _should not happen_ that case.
```
lp, err := fes.routingService.LookupCli()
	if err != nil {
		errCh <- fmt.Errorf("routing service cli not found: %v", err)
		return
	}
```

cmd/frontend/internal/frontend/service.go
```diff
@@ -288,95 +293,79 @@ func (fes *FrontEndService) Monitor(ctx context.Context, errCh chan<- error) {
...
+			if strings.Contains(protocolOut, bird.NoProtocolsLog) {
+				logger.Info("protocol output", "out", protocolOut)
+				denounce = true
+				continue
+			}
+
+			bfdOut, err := fes.routingService.ShowBfdSessions(ctx, lp, `NBR-BFD`)
+			if err != nil {
+				logger.Error(err, "BFD output", "out", strings.Split(bfdOut, "\n"))
+				denounce = true
```
zolug, line 312: Again, I don't think we should do denounce.

cmd/frontend/internal/frontend/service.go
```diff
@@ -258,25 +258,30 @@ func (fes *FrontEndService) Monitor(ctx context.Context, errCh chan<- error) {
...
+				hasConnectivity = false
+				if refreshCancel != nil {
+					refreshCancel()
+					refreshCancel = nil
+				}
+				if err := denounceFrontend(ctx, fes.targetRegistryClient); err != nil {
+					logger.Error(err, "denounceFrontend")
+				}
+				connectivityMap = map[string]bool{}
+				delay = 3 * time.Second // Avoid flapping (more?)
```
zolug, line 282: Not sure about this one. An env var would be nice.

cmd/frontend/internal/frontend/service.go
uablrek: Can WaitStart() be moved to before the go func()?

cmd/frontend/internal/frontend/service.go
```diff
@@ -258,25 +258,30 @@ func (fes *FrontEndService) Monitor(ctx context.Context, errCh chan<- error) {
...
+			}
+
 			select {
 			case <-ctx.Done():
 				logger.Info("Shutting down")
```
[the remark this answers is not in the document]
uablrek: Ok, I add that. But with a brand new context

cmd/frontend/internal/frontend/service.go
uablrek: ok

cmd/frontend/internal/frontend/service.go
```diff
@@ -288,95 +293,79 @@ func (fes *FrontEndService) Monitor(ctx context.Context, errCh chan<- error) {
...
-					if refreshCancel != nil {
-						refreshCancel()
-					}
-					_ = denounceFrontend(fes.targetRegistryClient)
-					noConnectivity = true
-					connectivityMap = map[string]bool{}
-					logger.Info("protocol output", "out", protocolOut)
-					//linkCh <- "No protocols match"
+				logger.Error(err, "protocol output", "out", strings.Split(protocolOut, "\n"))
+				denounce = true
```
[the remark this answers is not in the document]
uablrek: I will add a counter for the "session errors" and denounce after X tries (X configurable)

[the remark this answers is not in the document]
zolug: Btw, whenever the code deems denounce is required, the VIPs must be denounced as well. (Unless the FE is terminating/restarting).
Otherwise, for example if the connection is still UP according to bird, then BGP will still attract traffic.

Similarly, health.SetServingStatus should be adjusted.

uablrek:
Updated. I will merge and force-push after all is settled. I have not tested to set the env variables. I wait with that until I have fixed all review comments.
Also the announce log is temporary on DEBUG, I must remember to set it back to TRACE.
BTW is the 5s hard-coded timeout for announce/denounce ok?

cmd/frontend/internal/frontend/service.go
uablrek: You are right. I move them back into the go function.

cmd/frontend/internal/frontend/service.go
uablrek: Honestly I have no idea what those errors I called "session errors" are and what that can cause them. So, please take a close look, and suggest what to do differently if needed

cmd/frontend/internal/frontend/service.go
```diff
@@ -293,25 +335,25 @@ func (fes *FrontEndService) Monitor(ctx context.Context, errCh chan<- error) {
...
 				logger.Error(err, "protocol output", "out", strings.Split(protocolOut, "\n"))
-				denounce = true
+				sessionErrors++
 				continue
 			}
-
 			if strings.Contains(protocolOut, bird.NoProtocolsLog) {
 				logger.Info("protocol output", "out", protocolOut)
-				denounce = true
+				sessionErrors++
```
zolug, line 357: This is not a "session error". It indicates that there are no gateways configured. Either not added yet (in which case it doesn't matter if it was a session error or not), or all the gateways had been removed from the Attractor (FE config).
A denounce is required.

cmd/frontend/internal/frontend/service.go
zolug: That question is still in place. Would be better to remove, in order to avoid confusion.
Without the retry block, the FE would not be able to announce its connectivity without a prior state change (-> down, up).

cmd/frontend/internal/frontend/service.go
```diff
@@ -252,24 +252,66 @@ func (fes *FrontEndService) SetNewConfig(ctx context.Context, c interface{}) err
...
 	}
 
+	var (
+		// hasConnectivity indicates external connectivity
+		// requires 1 GW per IP family if configured to be reachable
+		hasConnectivity bool
+		// sessionErrors may be temporary
+		sessionErrors int
+		// maxSessionErrors is max sessionErrors before a denounce
+		maxSessionErrors int = 3
```
zolug, line 263: No clue why these errors pop up. But I would prefer a bit higher value, maybe 5.

cmd/frontend/internal/frontend/service.go
```diff
@@ -330,6 +333,12 @@ func (fes *FrontEndService) Monitor(ctx context.Context, errCh chan<- error) {
 				if refreshCancel != nil {
 					refreshCancel()
 				}
+				if hasConnectivity {
+					// (logging will not work inside denounceFrontend())
+					if err := denounceFrontend(context.Background(), fes.targetRegistryClient); err != nil {
```
zolug, line 339: I wonder if blocking the shutdown for 5 seconds could be a bit too much if e.g. NSP is not around.
Anyways, we can revisit it, if required...

cmd/frontend/internal/frontend/service.go
```diff
@@ -258,25 +258,30 @@ func (fes *FrontEndService) Monitor(ctx context.Context, errCh chan<- error) {
...
 		for {
+
+			if denounce {
+				denounce = false
+				hasConnectivity = false
+				if refreshCancel != nil {
+					refreshCancel()
+					refreshCancel = nil
+				}
+				if err := denounceFrontend(ctx, fes.targetRegistryClient); err != nil {
```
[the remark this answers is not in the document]
uablrek: > or some simple but lame init delay could be used.

I go for this option as a start. Same as before, a 3s delay

uablrek:
I added a pointer to the config struct in `FrontEndService struct`. IMO that's better than to copy all config vars. Not only unnecessary code is removed, but also that the value is actually a configurable becomes clear.

cmd/frontend/internal/frontend/service.go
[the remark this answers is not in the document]
zolug: Seems that in the baseline it wasn't the delay rather the grpc option WaitForReady that alowed the initial (or any) denounce to succeed. It blocks RPCs until the connection is ready.

Even with the 3 seconds delay, denounce fails on xcluster for me.

I think the easiest way forward would be if denounce did not have hard coded timeout, instead it would be the callers responsibility to pass a feasible context. Or maybe just increase the hard-coded value in denounce to 30 seconds, and in case of shutdown pass a context with lower timeout.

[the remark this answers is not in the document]
uablrek: I think it may be best just to remove the timeout in announceFrontend/denounceFrontend entirely and make a comment. Not adding a ctx with timeout at all in the calls. I.e. same as it was before. Timeouts would be hard to decide, especially if they are dynamic. The passed ctx can still be canceled from main.

Or do you think there is a case where the calls can hang for an unreasonable time (whatever that is)?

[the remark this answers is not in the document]
uablrek: > and in case of shutdown pass a context with lower timeout.

I missed this. Of course, a timeout is needed on shutdown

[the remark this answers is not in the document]
zolug: You're right. Expect for the shutdown, I don't see much need for timeout.

PR #421, opened 2023-05-26
Allow multi NAD in attractor

LionelJouin:
## Description

Multi NAD allows multiple network interfaces to be added to the attractor pods. The feature is already developed, only some conditions prevented its usage.
The interface name that will be injected in the frontend via the environment variable will be the one specified in the attractor specs

## Issue link

/

## Checklist

- Purpose
    - [ ] Bug fix
    - [x] New functionality
    - [ ] Documentation
    - [ ] Refactoring
    - [ ] CI
- Test
    - [ ] Unit test
    - [ ] E2E Test
    - [x] Tested manually
- Introduce a breaking change
    - [x] No

### DIFF
--- a/pkg/controllers/attractor/stateless-lb-frontend.go
+++ b/pkg/controllers/attractor/stateless-lb-frontend.go
@@ -90,22 +90,13 @@ func (l *LoadBalancer) getNscEnvVars(allEnv []corev1.EnvVar) []corev1.EnvVar {
 
 func (l *LoadBalancer) getFeEnvVars(allEnv []corev1.EnvVar) []corev1.EnvVar {
 	operatorEnv := map[string]string{
-		"NFE_CONFIG_MAP_NAME": common.ConfigMapName(l.trench),
-		"NFE_NSP_SERVICE":     common.NSPServiceWithPort(l.trench),
-		"NFE_TRENCH_NAME":     l.trench.ObjectMeta.Name,
-		"NFE_ATTRACTOR_NAME":  l.attractor.ObjectMeta.Name,
-		"NFE_NAMESPACE":       l.attractor.ObjectMeta.Namespace,
-		"NFE_EXTERNAL_INTERFACE": func() string {
-			externalInterface := l.attractor.Spec.Interface.Name
-			// if set use the interface provided by the Network Attachment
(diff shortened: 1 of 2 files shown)


pkg/controllers/attractor/stateless-lb-frontend.go
```diff
@@ -90,22 +90,13 @@ func (l *LoadBalancer) getNscEnvVars(allEnv []corev1.EnvVar) []corev1.EnvVar {
...
-			}
-			return externalInterface
-		}(),
-		"NFE_LOG_LEVEL": common.GetLogLevel(),
+		"NFE_CONFIG_MAP_NAME":    common.ConfigMapName(l.trench),
+		"NFE_NSP_SERVICE":        common.NSPServiceWithPort(l.trench),
+		"NFE_TRENCH_NAME":        l.trench.ObjectMeta.Name,
+		"NFE_ATTRACTOR_NAME":     l.attractor.ObjectMeta.Name,
+		"NFE_NAMESPACE":          l.attractor.ObjectMeta.Namespace,
+		"NFE_EXTERNAL_INTERFACE": l.attractor.Spec.Interface.Name,
```
zolug, line 98: If I remember right, when the code appends the annotations, it interface name to be used in the POD is decided based on NAD config in the Attractor. Meaning, if there's no interface name in the NAD, the code will derive the name from the Specs.Interface.Name.
So, based on this change, the "primary" Attractor NAD supplying the external interface must not have an Interface Name set.
Also, in case multiple NADs are set, but none of them have an interface name specified, the code would try to re-use the same Specs.Interface.Name as POD interface names when appending the NADs to the resource template, wouldn't it?

LionelJouin: Yes, the user should give a specific interface name in the non-primary network-attachments in the attractor spec. The primary one should not have any name (or it should match with the .Specs.Interface.Name).

There is an example here: https://gist.github.com/LionelJouin/3cf494db1db592fb01b16ff16b6493c8
The attractor has 2 NADs, and the macvlan is a secondary one, the primary one is the vlan-100 that will get the ext-vlan0 name.

PR #442, opened 2023-07-13
FE; make BIRD use IANA approved ports for BFD (49152-65535)

zolug:
## Description
As per BIRD documentation, use ip_local_port_range to force BIRD using IANA approved src ports for BFD.

```BFD packets are sent with a dynamic source port number. Linux systems use by default a bit different dynamic port range than the IANA approved one (49152-65535). If you experience problems with compatibility, please adjust /proc/sys/net/ipv4/ip_local_port_range.```

## Issue link
NA

## Checklist

- Purpose
    - [x] Bug fix
    - [ ] New functionality
    - [ ] Documentation
    - [ ] Refactoring
    - [ ] CI
- Test
    - [ ] Unit test
    - [ ] E2E Test
    - [x] Tested manually
- Introduce a breaking change
    - [x] No

### DIFF
--- a/deployments/helm/templates/_helpers.tpl
+++ b/deployments/helm/templates/_helpers.tpl
@@ -6,11 +6,11 @@ Set IP Family
 
 {{- define "meridio.loadBalancer.sysctls" -}}
 {{- if eq .Values.ipFamily "dualstack" -}}
-{{- printf "sysctl -w net.ipv6.conf.all.forwarding=1 ; sysctl -w net.ipv4.conf.all.forwarding=1 ; sysctl -w net.ipv4.fib_multipath_hash_policy=1 ; sysctl -w net.ipv6.fib_multipath_hash_policy=1 ; sysctl -w net.ipv4.conf.all.rp_filter=0 ; sysctl -w net.ipv4.conf.default.rp_filter=0" -}}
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: pkg/controllers/attractor/stateless-lb-frontend.go [hunk 1]
```diff
@@ -202,7 +202,7 @@ func (l *LoadBalancer) insertParameters(dep *appsv1.Deployment) *appsv1.Deployme
 	}
 	ret.Spec.Template.Spec.InitContainers[0].Args = []string{
 		"-c",
-		common.GetLoadBalancerSysCtl(l.trench),
+		fmt.Sprintf("%s ; %s", common.GetLoadBalancerSysCtl(l.trench), common.GetFrontendSysCtl()),
```
LionelJouin, line 205: Could be added here otherwise: https://github.com/Nordix/Meridio/blob/v1.0.9/pkg/controllers/common/models.go#L42

LionelJouin (commented):
Just for reference for IPv6: https://tldp.org/HOWTO/Linux+IPv6-HOWTO/ch11s03.html

Also, could you update the documentation: https://github.com/Nordix/Meridio/blob/v1.0.9/docs/components/frontend.md?plain=1#L98
```
Sysctl: net.ipv4.ip_local_port_range='49152 65535' | The source port of BFD Control packets must be in the range 49152-65535
```

pkg/controllers/attractor/stateless-lb-frontend.go
zolug: Yeah, but I wanted to handle it separately, even though currently LB and FE share the same POD, the settings are FE specific.
There were plans to support a non-composite setups in the future, in which case I figured separating these could help sort things out.

LionelJouin (approved):

Issue #419, opened 2023-05-23
Collect Metrics


LionelJouin:
**Is your feature request related to a problem? Please describe.**

Add metrics to Meridio to improve observability tools.
Here is some slides:
https://docs.google.com/presentation/d/1yuiDj7H4NZTea7dJAKPK4SBvkHtZyWrn5HisNe1shuI
And OpenTelemetry/Prometheus/Grafana stack deployment instruction:
https://gist.github.com/LionelJouin/cfa15a569f1f23d8a84d43dc73b5f373

**Describe the solution you'd like**

## Interfaces in stateless-lb-frontend / Proxy / TAPA

`<interface.metric>`: rx_packets, tx_packets, rx_bytes, tx_bytes, rx_errors, tx_errors, rx_dropped, tx_dropped

- Name: meridio.interface.`<interface.metric>`
- Description: `<interface.metric>` metrics for the network interface
- Type: Counter
- Value: `<interface.metric>`
- Attributes:
   - Pod Name
   - Trench
   - Conduit (optional)
   - Attractor (optional)
   - Interface Name

## Interfaces in stateless-lb-frontend / Proxy / TAPA ?

- Name: meridio.interface.status
- Description: Network interface status
- Type: Gauge (Health Metric)
- Value: Status of the interface (1 if up, 0 if down)
- Attributes:
   - Pod Name
   - Trench
   - Conduit (optional)
   - Attractor (optional)
   - Interface Name

## Stream status in conduit instance

- Name: meridio.conduit.stream.status
- Description: Stream status in the conduit instance
- Type: Gauge (Health Metric)
- Value: Status of the stream (1 if configured)
- Attributes:
   - Pod Name
   - Trench
   - Conduit
   - Stream

## Flows configured in conduit instance

- Name: meridio.conduit.stream.flow.status
- Description: Flow status in the conduit instance
- Type: Gauge (Health Metric)
- Value: Status of the flow (1 if configured) (nfqlb matches_count instead?)
- Attributes:
   - Pod Name
   - Trench
   - Conduit
   - Stream
   - Flow

## Targets configured in conduit instance

- Name: meridio.conduit.stream.target.status
- Description: Target status in the conduit instance
- Type: Gauge (Health Metric)
- Value: Status of the target (1 if configured, 0 if pending)
- Attributes:
   - Pod Name
   - Trench
   - Conduit
   - Stream
   - Target (identifier + IPs)

## Targets configured in conduit instance

[...]

PR #477, opened 2023-11-07
Gateway routes imported/exported/preferred metrics

LionelJouin:
## Description

Collect metrics of routes imported, exported and preferred for each gateways. The metrics are extracted and parsed from show protocol all command in bird.

## Issue link

https://github.com/Nordix/Meridio/issues/419

## Checklist

- Purpose
    - [ ] Bug fix
    - [x] New functionality
    - [ ] Documentation
    - [ ] Refactoring
    - [ ] CI
- Test
    - [ ] Unit test
    - [ ] E2E Test
    - [x] Tested manually
- Introduce a breaking change
    - [x] No

### DIFF
--- a/cmd/frontend/internal/frontend/metrics.go
+++ b/cmd/frontend/internal/frontend/metrics.go
@@ -1,0 +1,260 @@
+/*
+Copyright (c) 2023 Nordix Foundation
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package frontend
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"regexp"
+	"strconv"
+	"strings"
+	"sync"
+
+	nspAPI "github.com/nordix/meridio/api/nsp/v1"
+	"github.com/nordix/meridio/cmd/frontend/internal/bird"
+	meridioMetrics "github.com/nordix/meridio/pkg/metrics"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+)
+
+type GatewayMetrics struct {
+	meter            metric.Meter
+	gateways         map[string]*nspAPI.Gateway
+	metricAttributes []metric.ObserveOption
+	RoutingService   *bird.RoutingService
+	mu               sync.Mutex
+}
+
+func NewGatewayMetrics(metricAttributes []metric.ObserveOption) *GatewayMetrics {
+	meter := otel.GetMeterProvider().Meter(meridioMetrics.METER_NAME)
+	gm := &GatewayMetrics{
+		meter:            meter,
+		metricAttributes: metricAttributes,
+	}
+
+	return gm
+}
+
+func (gm *GatewayMetrics) Set(gateways []*nspAPI.Gateway) {
+	gatewaysMap := map[string]*nspAPI.Gateway{}
+
+	for _, gateway := range gateways {
+		gatewaysMap[fmt.Sprintf("NBR-%s", gateway.Name)] = gateway
+	}
+
+	gm.mu.Lock()
+	defer gm.mu.Unlock()
+	gm.gateways = gatewaysMap
+}
+
+// Collect collects the metrics for the gateways.
+func (gm *GatewayMetrics) Collect() error {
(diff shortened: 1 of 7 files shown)


LionelJouin:
Here is a small experiment regarding cpu consumption with 2 exported routes and 1 imported.
`watch -n 0.1 curl -k https://localhost:2224/metrics`
Idle: 10m
1 req/sec: 15m
10 req/sec: 60m

cmd/frontend/internal/frontend/metrics.go
```diff
@@ -0,0 +1,260 @@
...
+func (bs *BirdStats) String() string {
+	return fmt.Sprintf("%s (%d, %d, %d)", bs.gatewayName, bs.routesImported, bs.routesExported, bs.routesPreferred)
+}
+
+func ParseShowProtocolsAll(output string) []*BirdStats {
+	res := []*BirdStats{}
+
+	protocols := strings.Split(output, "\n\n")
+
+	regex := regexp.MustCompile(`(?m)    Routes:.*`)
```
zolug, line 199: Because the regex does not change, performance wise it better to create a file local variable instead of compiling it on every run.

cmd/frontend/internal/frontend/metrics.go
```diff
@@ -0,0 +1,260 @@
...
+		}
+
+		newStat := &BirdStats{
+			gatewayName:     protocolName[0],
+			routesImported:  0,
+			routesExported:  0,
+			routesPreferred: 0,
+		}
+
+		routes := regex.FindAllString(protocol, -1)
```
zolug, line 215: Seems like a lot of hassle with all those string operations that follow.
Wouldn't a regex using group matching be easier: https://regex101.com/r/SV8ouF/1
Then you could rely on the single map approach from here: https://stackoverflow.com/questions/60109288/regex-named-groups-in-golang

cmd/frontend/internal/frontend/metrics.go
```diff
@@ -0,0 +1,260 @@
...
+		return fmt.Errorf("frontend metrics, failed to Int64ObservableCounter (%s): %w", meridioMetrics.MERIDIO_ATTRACTOR_GATEWAY_ROUTES_PREFERRED, err)
+	}
+	return nil
+}
+
+func (gm *GatewayMetrics) observe(ctx context.Context, observer metric.Int64Observer, valueFunc func(*BirdStats) int64, lp string) error {
+	gm.mu.Lock()
+	defer gm.mu.Unlock()
+
+	spaOutput, err := gm.RoutingService.ShowProtocolSessions(ctx, lp, "")
```
zolug, line 140: This captures all the things bird handles as protocol, yet only the ones with the prefix "NBR-" are of interest.
e.g.: ```ShowProtocolSessions(ctx, lp, `NBR-*`)```

cmd/frontend/internal/frontend/metrics.go
```diff
@@ -0,0 +1,260 @@
...
+	_, err = gm.meter.Int64ObservableCounter(
+		meridioMetrics.MERIDIO_ATTRACTOR_GATEWAY_ROUTES_IMPORTED,
+		metric.WithUnit("routes"),
+		metric.WithDescription("Counts number of routes imported for a gateway."),
+		metric.WithInt64Callback(func(ctx context.Context, observer metric.Int64Observer) error {
+			return gm.observe(
+				ctx,
+				observer,
+				func(birdStats *BirdStats) int64 {
+					return int64(birdStats.routesImported)
```
zolug, line 86: Feels clumsy that the same operation has to be repeated 3 times in the background because 3 values must be extracted. Isn't there some recommendation in opentelemtry how to avoid this?

zolug:
> Here is a small experiment regarding cpu consumption with 2 exported routes and 1 imported. `watch -n 0.1 curl -k https://localhost:2224/metrics` Idle: 10m 1 req/sec: 15m 10 req/sec: 60m

Btw. with 100 000 IPv4 routes (net prefix /32) in Kind on my machine:
idle: 10m
1 req/sec: 20m
10 req/sec: 100m

LionelJouin:
> > Here is a small experiment regarding cpu consumption with 2 exported routes and 1 imported. `watch -n 0.1 curl -k https://localhost:2224/metrics` Idle: 10m 1 req/sec: 15m 10 req/sec: 60m
> 
> Btw. with 100 000 IPv4 routes (net prefix /32) in Kind on my machine: idle: 10m 1 req/sec: 20m 10 req/sec: 100m

Thanks for checking this, it seems acceptable. Usually Prometheus should fetch the metrics so often.

cmd/frontend/internal/frontend/metrics.go
LionelJouin: I don't know, I find it easier to maintain and troubleshoot this than a complex regex

cmd/frontend/internal/frontend/metrics.go
LionelJouin: No, I don't think opentelemtry gives any recommendation. they are different types of metrics, like rx/tx bytes/packets for an interface.

PR #490, opened 2023-12-14
LB logging and some improvements

zolug:
## Description
Improved logs to help troubleshooting.

other:
    - Monitor NSM connections the LB is involved in for DELETE
      and DOWN events for logging (functionality shared by lb and proxy)
    - Stream lb instance unsubscribes at interface monitor on delete
    - LB subscribes to native interface monitor to learn about
      interface removal to sync its interface list
    - Monitor Target reachability based on kernel neighbor events,
      and log changes

## Issue link
NA

## Checklist

- Purpose
    - [x] Bug fix
    - [ ] New functionality
    - [ ] Documentation
    - [ ] Refactoring
    - [ ] CI
- Test
    - [ ] Unit test
    - [ ] E2E Test
    - [x] Tested manually
- Introduce a breaking change
    - [x] No

### DIFF
--- a/pkg/kernel/neighbor/monitor.go
+++ b/pkg/kernel/neighbor/monitor.go
@@ -1,0 +1,133 @@
+/*
+Copyright (c) 2024 Nordix Foundation
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package neighbor
+
+import (
+	"context"
+	"sync"
+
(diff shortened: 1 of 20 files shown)


pkg/kernel/neighbor/monitor.go
```diff
@@ -0,0 +1,134 @@
+/*
```
LionelJouin, line 1: I guess it's ok, but I would still prefer to have metrics for with the IP / MAC as attribute and the state as value than having the neighbor table added to the logs.

LionelJouin (approved):

pkg/kernel/neighbor/monitor.go
zolug: Proper metrics for Targets would be nice indeed.
(Though currently I think MAC might change even without unregistering the target first. Also, MAC is not known until successfully resolved, meaning the key won't be complete from start.)

pkg/networking/ifaces.go
```diff
@@ -88,3 +92,13 @@ type InterfaceMonitorSubscriber interface {
 	InterfaceCreated(Iface)
 	InterfaceDeleted(Iface)
 }
+
+type NeighborMonitor interface {
```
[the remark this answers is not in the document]
zolug: Alright, I thought it's a common place holding networking related interfaces.

Issue #500, opened 2024-02-13
TAPA mem leak during periodic Stream Close-Open


zolug:
**Describe the bug**
Consider the following resource settings for TAPA container:
```
       resources:
            limits:
              cpu:     64m
              memory:  48Mi
            requests:
              cpu:     32m
              memory:  24Mi
```
If TAPA keeps closing and re-opening the single Stream it was connected to from start, the memory footprint increases. And eventually leads to OOM kill.

**To Reproduce**
1. Deploy and setup Prometheus to monitor the memory usage of example-target POD's TAPA container.
2. Deploy 1 Trench with 1 Conduit, 1 Attractor, 1 Stream and 1 Flow.
3. Deploy example-target with the resource settings above, and let it open the single Steam configured.
4. Open a dashboard and pick a target-example POD instance to monitor its memory usage.
5. With a simple script open the Stream and after 10 seconds Close it in a loop for the chosen target-example POD. E.g.:
```
while [ 1 ]; do date; kc exec target-a-54df657d88-55h8m -n nvip -- ./target-client open -t trench-a -c load-balancer-a1 -s stream-a1; sleep 10;kc exec target-a-54df657d88-55h8m -n nvip -- ./target-client close -t trench-a -c load-balancer-a1 -s stream-a1;echo ""; done
```

**Expected behavior**
Should not leak.

**Context**
 - Kubernetes: v1.26.6
 - Network Service Mesh: v1.12.0
 - Meridio: commit 7e2669ce372761e500f426c5b6983696015f4016 (HEAD -> master, origin/master, origin/HEAD)
Author: Lugossy Zoltan <zoltan.lugossy@est.tech>
Date:   Tue Feb 6 15:40:29 2024 +0100

    metrics; fix collecting flow stats

**Logs**
NA

zolug:
There seems to be a goroutine leak in interfacename.InterfaceNameChache:
- ReleaseTrigger() usage is faulty, it started goroutine is not cancelled if the release gets aborted
- There's also no point calling releaseTrigger() a second time in pendingRelease() if the inteface name couldn't be cancelled immediately in Release().

This problem might arise, if NSM is trying to heal a TAPA connection. The longer it takes to connect with a Proxy, the more goroutines are created. If I'm not mistaken, these goroutines should eventually exit (after 10 minutes or so), but they increase memory usage till then.

zolug:
[...]

PR #503, opened 2024-02-16
TAPA; fix goroutine cleanup in intefacename cache

zolug:
## Description
Stop releaseTrigger() invoked goroutine(s) right away when Release() is cancelled in InterfaceNameChache.
Thus, avoid accumulation of such goroutines (leading to increased memory footprint) for example when TAPA must keep trying to establish connection with the local Proxy for longer time.

## Issue link
#500 

## Checklist

- Purpose
    - [x] Bug fix
    - [ ] New functionality
    - [ ] Documentation
    - [ ] Refactoring
    - [ ] CI
- Test
    - [x] Unit test
    - [ ] E2E Test
    - [x] Tested manually
- Introduce a breaking change
    - [x] No

### DIFF
--- a/pkg/nsm/interfacename/cache.go
+++ b/pkg/nsm/interfacename/cache.go
@@ -28,7 +28,7 @@ import (
 
 const defaultReleaseTimeout = 600 * time.Second
 
-type ReleaseTrigger func() <-chan struct{}
+type ReleaseTrigger func(ctx context.Context) <-chan struct{}
 
 // InterfaceNameChache -
 // InterfaceNameChache keeps track of interface names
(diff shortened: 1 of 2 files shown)


LionelJouin (approved):
The interface name generator became too complex. Removing the chain element and setting the name as part of the initial request (in ambassador conduit) could be an option?

zolug:
> The interface name generator became too complex. Removing the chain element and setting the name as part of the initial request (in ambassador conduit) could be an option?

That's where we moved from towards this cache approach, because NSM failed to handle interface name updates for an existing connection (fetched from nsmgr via the monitor connection API). And there were all sort of problems with policy based routes as well.
Some of the problems might have been already fixed, but I haven't checked it. And some older NSM versions that we still are required to support might be still affected by the problem. Eventually we should re-test NSM without this interfacename cache, but this PR isn't really adding to complexity otherwise.

Issue #521, opened 2024-04-22
Proxy releases NSM connection IPs too early


zolug:
**Describe the bug**
Proxy might release the IP addresses allocated to NSM connection via the IPAM too early.

There's a custom NSM chain element that hooks into Request and Close calls to allocate or release IP addresses for NSM connections the Proxy is involved in. Main problem is, that the IPs are released right when Close() gets invoked.

This means that there's at least a short time window, when the IPs are still in use, yet they can be re-assigned to some other NSM connection.

Although, a much severe problem can arise when the proxy acts as an NSC (i.e. connects LBs), and the Close call fails. In which case the old interface remains in the POD with the addresses that are considered ready to be leased.
This might for example happen due to local NSMgr restart, when NSM heal activates. If data-plane monitoring is not enabled (which is not in our case), then heal tries to do a reconnect part of which it first tries to Close the connections. Close might fail if NSMgr restart have not concluded yet leaving the network interfaces intact. While during a subsequent Request (as part of heal) IPs might get re-allocated to different NSC connections. This might lead to duplicated IP issue where 2 NSM interfaces are using the same IP address.

It has been seen that on 1 cluster NSM heal failed to heal 1 of the 2 connections towards LBs in a Proxy after an IP swap described above. In which case the duplicated IP problem persisted. The NSM connection Requests related to the connection not able to get healed did reach the local vpp-forwarder but did not progress beyond beginServer.Request().(No errors or anything else were visible. The Requests thus timed out in the local nsmgr.)

**To Reproduce**
Did not manage to reproduce. But was seen after NSMgr restart on a system with 2 LB-FEs.

**Expected behavior**
IPs must not get re-assigned to other connections while they are still in use.

**Context**
 - Network Service Mesh: v1.11.2
 - Meridio: v1.0.15
 ...

**Logs**
Add logs here.

zolug:
It has been decided that as a short-term solution IP release shall be delayed in the Proxy.

While the long-term solution could be to rework the IPAM solution so that IPs are not allocated for eternity. Instead there should be an option to lease them for a certain amount of time.
For this, the db model should be extended to support expiration time like features. NSM connections are refreshed periodically. During each refresh the IPs allocation is verified, which could be used to update the db entry in IPAM. A garbage collector could be introduced to reap db entries with lifetime who haven't prolonged their leases.
[...]

PR #524, opened 2024-04-26
Delayed IP release of NSM connection IPs

zolug:
## Description
It's really hard to asses when IP addresses allocated to NSM connections are no longer needed,
without risking early re-allocation that could result in duplicated IP issues across the NSM interfaces.

For example NSM Close() might fail, but even if it succeeds there's no guarantee that the NSM interfaces associated
with the connection are actually removed. (After nsmgr restart the new instance will not return an error for a Close() request the connection is not known to it.)

As a short-term fix, IP release is delayed by a configurable value. The main goal is to avoid freeing up IPs of NSM connections subject to NSM heal even if the related Close/Request calls might timeout. The default delay value is set to exceed the current timeout value of NSM requests (15s).

The delay can be configured via the Operator environment variable PROXY_IP_RELEASE_DELAY which in turn configures a respective env variable for the proxy containers.

## Issue link
#521

## Checklist

- Purpose
    - [x] Bug fix
    - [ ] New functionality
    - [ ] Documentation
    - [ ] Refactoring
    - [ ] CI
- Test
    - [ ] Unit test
    - [ ] E2E Test
    - [x] Tested manually
- Introduce a breaking change
    - [x] No


pkg/proxy/proxy.go
```diff
@@ -548,6 +570,56 @@ func (p *Proxy) SetVIPs(vips []string) {
...
+	for id, cancel := range p.connectionToReleaseMap {
+		cancel()
+		go func(ctx context.Context, id string) {
+			defer wg.Done()
+
+			_ = retry.Do(func() error {
+				ctxRelease, cancelRelease := context.WithTimeout(ctx, 4*time.Second)
+				defer cancelRelease()
+				for _, subnet := range p.Subnets {
+					child := &ipamAPI.Child{
```
LionelJouin, line 593: Repetition with l468

LionelJouin (approved):

Issue #527, opened 2024-05-02
IPAM improvements


zolug:
**Is your feature request related to a problem? Please describe.**
Follow-up item for: https://github.com/Nordix/Meridio/issues/521

Main problem is, that Proxies who are in charge of IP allocation/release might disappear abruptly and along with it so can NSM connections they hosted providing no second chance to release the allocated IPs. (Same can happen during planned termination of Proxy if IPAM POD is not available.)

There should be an option to lease IPs for a certain amount of time. This could secure that IP allocations are not leaked.
For this, the db model should be extended to support expiration time like features.

NSM connections are refreshed periodically. During each refresh the IP allocation is verified, which could be used to update the db entry in IPAM. A garbage collector could be introduced to reap db entries with lifetime who haven't prolonged their leases and thus expired.

Upon upgrade, the new IPAM instance should migrate the existing database to introduce expiration time support for old db entries that belong to NSM connections. (Obviously rollback could result in serious disturbances.)

**Describe the solution you'd like**
See above.

**Describe alternatives you've considered**
NA

**Additional context**
NA

PR #532, opened 2024-06-26
ipam; introduce garbage collection logic

zolug:
## Description
Introduced garbage collection logic in IPAM to clean up IP records not updated for a long time and hence considered not in use anymore by any NSM connections.

For this the database model has been extended with UpdatedAt (timestamp) and Expirable (*bool) fields. UpdatedAt is used by newly introduced garbage collector logic along with a threshold time to determine if a record can be removed. While Expirable limits the scope of database records the garbage collector mechanism applies to. The check is run periodically. Both the threshold and periodicity interval are configurable through environment variables (IPAM_GARBAGE_COLLECTION_THRESHOLD and IPAM_GARBAGE_COLLECTION_INTERVAL).

As part of the changes node allocations do update existing database records to keep UpdatedAt values up to date.

The logic takes only records that belong to an NSM connection into consideration. This is achieved by means of applying a filter matching records whose Expirable field is true denoting NSM connection related entries. (For other allocations the IPAM is not consulted periodically.)

The model update re-uses the old database table extending it with UpdatedAt and Expirable columns. In case of rollback the excess columns are simply ignored by the "framework" without causing any harm.
When moving from a version using the old model to the new model, then automigrate won't set UpdatedAt values for existing records. However, a default false value is configured for Expirable which will be set for all records after automigrate. A custom migrate function takes care of the update of existing connection records setting their Expirable and UpdatedAt values, allowing them to be subject to the garbage collection mechanism.

In order to limit API changes the value of Expirable field is signalled via the context. Also, to avoid upgrade issues IPAM
gRPC proto haven't been extended (to allow IPAM users to chose if their allocation could expire).

The garbage collection (by default enabled) can be disabled through environment variable IPAM_GARBAGE_COLLECTION_ENABLED.

Notes:
IPAM_GARBAGE_COLLECTION_INTERVAL also determines when the garbage collector will be first invoked. Thus, setting it to a low value must be avoided to allow existing NSM connections to get their IP allocations refreshed as part of the recurring NSM connection refresh which is tied to NSM's MaxTokenLifetime (e.g. during IPAM upgrade/failover).
IPAM_GARBAGE_COLLECTION_THRESHOLD must be higher than NSM's MaxTokenLifetime.

## Issue link
#527 

## Checklist

- Purpose
    - [ ] Bug fix
    - [x] New functionality
    - [ ] Documentation
    - [ ] Refactoring
    - [ ] CI
- Test
    - [ ] Unit test
    - [ ] E2E Test
    - [x] Tested manually
- Introduce a breaking change
    - [x] No

### DIFF
--- a/pkg/ipam/storage/sqlite/sqlite.go
+++ b/pkg/ipam/storage/sqlite/sqlite.go
@@ -1,5 +1,6 @@
 /*
 Copyright (c) 2021 Nordix Foundation
+Copyright (c) 2024 OpenInfra Foundation Europe
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -21,13 +22,17 @@ import (
 	"errors"
 	"fmt"
 	"sync"
+	"time"
 
 	"github.com/nordix/meridio/pkg/ipam/prefix"
 	"github.com/nordix/meridio/pkg/ipam/types"
+	"github.com/nordix/meridio/pkg/log"
 	"gorm.io/driver/sqlite"
 	"gorm.io/gorm"
 	"gorm.io/gorm/logger"
 )
+
+const bridgeName = "bridge" // prefix with name "bridge" is special because it's not periodically updated
 
 type SQLiteIPAMStorage struct {
 	DB *gorm.DB
@@ -36,7 +41,8 @@ type SQLiteIPAMStorage struct {
 
 func New(datastore string) (*SQLiteIPAMStorage, error) {
 	db, err := gorm.Open(sqlite.Open(datastore), &gorm.Config{
-		Logger: logger.Default.LogMode(logger.Silent),
+		Logger:  logger.Default.LogMode(logger.Silent),
+		NowFunc: func() time.Time { return time.Now().UTC() },
 	})
 	if err != nil {
 		return nil, fmt.Errorf("failed to open db session: %w", err)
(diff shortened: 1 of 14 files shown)


pkg/ipam/storage/sqlite/sqlite.go
```diff
@@ -51,6 +58,197 @@ func New(datastore string) (*SQLiteIPAMStorage, error) {
...
+					log.Logger.Info("Garbage collector returned error", "err", err)
+				}
+			}
+		}
+	}()
+}
+
+// Fetch and delete -src/-dst records whose updatedAt timestamp is considered
+// expired according to the threshold.
+func (sqlis *SQLiteIPAMStorage) garbageCollector(ctx context.Context, threshold time.Duration) error {
```
LionelJouin, line 107: both `oneTimeGarbageCollector` and `garbageCollector` might be the same function except the sql  query

LionelJouin (commented):
I don't really like relying on the src and dst prefixes as nothing in the API is saying about the usage of those.
I guess if a node or a trench have their name finishing by -src or -dst, problem could happen.

zolug:
> I don't really like relying on the src and dst prefixes as nothing in the API is saying about the usage of those. I guess if a node or a trench have their name finishing by -src or -dst, problem could happen.

I do agree that the current approach is somewhat ugly and can be problematic.
My goal was to come up with a match criteria that does not require altering any of the existing fields, and could work also for entries inserted by older models. But probably the risk is way too high of removing other records.
 
Btw in case of trenches names, fortunately there could be no such problem as accidentally matching them, since ipam server already modifies their name based on the IP family type (adding either -0 or -1 to the trench name). Otherwise, for workers and conduits it would be definitely a problem if they would end with either -src or -dst as you pointed out.
 
 
some ideas how to move forward:
 
-Change IP allocation for other "IP resource types" such as trenches/conduits etc. to be recurring and hence updates could be utilized. (Therefore postfix matching could be completely ignored, and GC could be more generic.)
This might introduce a lot of changes and might be tricky to implement (haven't really looked into it). I guess it would be the best, if these updates could be done as part of ipamServer Allocate() (existence of all these are checked anyways here, but an idle system without NSM connections could be a problem).
 
-Add a new table that would indicate if an entry is "expirable", and GC would only consider those. In which case GC would not be able to clean-up leaked entries inserted according to old model.
Passing the information about an entry if its expirable could involve a lot of changes (inlcluding the API). Unless for example that info would be stored in the context (which normally is not a recommended usage of context yet happens often e.g. in NSM code).
 
-Trenches have no parents. Thefore use this knowledge to identify threnches, then fetch entries with trench parents thus getting the conduits. Then, find the conduit workers as those have conduits as parents. After that, we should have a list of trenches,conduits,conduit-workers, which could be used by GC to ignore records happaned to match anything in our list.
This obsiously exploits the hierarchy of entries, but it could be used to remove old leaked records created based on the old model.
Combining this with the "expirable" field would mean, that only the one-time GC should do all this "lookups" once.
 
-Rely on network prefix length. Ugly, because the storage doesn't know anything about such. But it would be an easy check for the GC to only remove entries related to /128 or /32 prefixes.
(Proxy bridges also have /32 and /128 but their name is hardcoded to bridge...)

LionelJouin (approved):

Issue #548, opened 2024-10-18
NSM interface replacement in TAPA can lead to 10 seconds traffic disturbance


zolug:
**Describe the bug**

**To Reproduce**
1. ...
2. ...

**Expected behavior**

**Context**
 - Kernel: [e.g. 5.15.43]
 - Kubernetes: [e.g. v1.24.1]
 - Spire: [e.g. v1.3.0]
 - Network Service Mesh: [e.g. v1.3.1]
 - Meridio: [e.g. v0.5.1]
 ...

**Logs**
Add logs here.

zolug:
ideas to resolve:

1., Lowering probe delay by tweaking sysctl params in LB, primarily ucast_solicit and delay_first_probe_time values:
Unfortunately, for the 'default' interface these two cannot be set in other than the init linux network namespace. Therefore, we cannot employ a privileged init container or a tuning network-attachment-definition. Therefore, in order to work these values should be set per NSM interface instead by some privileged entity.

2., Use arping program in TAPA (NSC) to send out gARP:
It would require CAP_RAW privilege at least. Any privilege requirement towards TAPA users must be avoided. Could be outsourced to NSM.

3., Change NSM Target registration protocol. E.g. report if associated NSM connection's state becomes DELETE, upon which LBs could remove any cached neighbor entries.

4., Proxy to send promisc gARP instead of the TAPA when a new NSM connection gets established between a TAPA and said proxy:
Proxy would need to learn the MAC of the NSM interface in the TAPA. For which it could use some ping or anything. Once, the MAC was learned, it could do a gARP on behalf of the TAPA using arping program with promisc mode. Due to the promisc mode, most probably it would require additional privileges than just CAP_RAW (needed to create RAW socket to send out gARP).

5., Using some custom communication channel proxy could signal the Target IPs to LBs when a new TAPA->proxy connection gets established. So, that the LB could remove any possible cached neighbor entries.

6., LBs could rely on NSM's Connection Monitor feature to learn if TAPA->Proxy NSM connection state changes to DELETE, and extract associated IPs from the connection to issue neighbor delete operations. Unfortunately, MonitorScopeSelector in NSM currently only supports PathSegment related filters like path name, id and token. Which are not known by LB, and can change dynamically over time. On the other hand, without filters there's a risk of receiving way too much events depending on the deployments using NSM.

zolug:
NSM MonitorScopeSelector should from now on support network service based filtering:
[...]

PR #549, opened 2024-12-11
LB to remove invalid Target neighbor entries

zolug:
## Description
LB applies cluster-wide monitoring of NSM connection Delete events between TAPA and Proxy, to remove invalid Target entries from the Linux neighbor cache preventing connection disturbances.

When TAPA experiences a NSM Heal event for example due to an unexpected connection refresh failure, the NSM connection is first closed and the associated interface is removed. Thus, the MAC address associated with the Target IP(s)
is no longer valid. Even if the same Target IP(s) will get re-assigned when the Heal process concludes, it's unlikely that the same MAC address will be assigned to the new NSM interface.

## Issue link
#548 

## Checklist

- Purpose
    - [x] Bug fix
    - [x] New functionality
    - [ ] Documentation
    - [ ] Refactoring
    - [ ] CI
- Test
    - [ ] Unit test
    - [ ] E2E Test
    - [x] Tested manually
- Introduce a breaking change
    - [x] No


zolug:
TODO: CI image requires golang update to 1.23

zolug:
TODO: replace v1.14.2-rc.1 with v1.14.2 once released

zolug:
rebased on master

Issue #555, opened 2025-02-20
Support for sequential Conduit updates


zolug:
**Is your feature request related to a problem? Please describe.**
When multiple Conduits are updated at once (for example part of a version upgrade) it might be desirable to serialize these updates so that only 1 Conduit would be affected at a time. This could be beneficial for example if 2 Conduits are used to accomplish a communication model with a "primary" and a "secondary" fallback path.

**Describe the solution you'd like**
The feature should be implemented in the Operator. Conduit Custom Resources could be optionally grouped based on the value of a dedicated annotation. By using a locking mechanism, the Operator could ensure that only one Conduit from the same group processes updates at a time.

**Describe alternatives you've considered**
The Conduit Custom Resource mostly translates to a DaemonSet implementing the Proxy logic. There seems to be no kubernetes native way of coordinating updates between separate DaemonSets.

PR #556, opened 2025-02-20
operator; enable sequential conduit updates

zolug:
## Description
Users can group Conduits into update groups by annotating Conduit Custom Resources with the `update-sync-group` key. Within an update group, Conduit updates are executed sequentially rather than in parallel.

The default annotation key can be changed via the operator environment variable `CONDUIT_UPDATE_SYNC_GROUP_KEY` when deploying the operator.

Locking is achieved using in-memory locks, which are fast and provide a sufficiently good solution. However, this approach is not effective in case of operator crashes during ongoing updates. Such crashes would be considered a bug to be fixed and should not occur in general.

The feature has support for changing the annotation value of a Conduit Custom Resource on the fly. (Including removing it completely.)
However, no immediate control can be expected when changing the update group of a Conduit whose Proxy's Status subresource still indicates underlying changes (e.g. in POD readiness) while the DaemonSet otherwise reflects the desired state. This is because the logic relies on the Status subresource to accurately track the progress of updates and ensure consistency.

Also, existing Conduit Custom Resources can be annotated before upgrading to a Meridio version that supports serialized Conduit updates. In this case, the new operator will respect the update groups of the "old" Conduits from the start.

Note: There is no benefit to having a single Conduit in an update group.

## Issue link
#555 

## Checklist

- Purpose
    - [ ] Bug fix
    - [x] New functionality
    - [ ] Documentation
    - [ ] Refactoring
    - [ ] CI
- Test
    - [ ] Unit test
    - [ ] E2E Test
    - [x] Tested manually
- Introduce a breaking change
    - [x] No

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -132,7 +132,7 @@ func main() {
 			"Enabling this will ensure there is only one active controller manager.")
 
 	if os.Getenv(common.LogLevelEnv) == "" { // trace as default value
-		os.Setenv(common.LogLevelEnv, "trace")
+		os.Setenv(common.LogLevelEnv, "TRACE")
 	}
 
 	ver := flag.Bool("version", false, "Print version and quit")
@@ -155,7 +155,7 @@ func main() {
 		os.Exit(0)
 	}
 
-	logger := log.New("Operator", common.LogLevelEnv)
+	logger := log.New("Operator", os.Getenv(common.LogLevelEnv))
 	ctrl.SetLogger(logger)
 
 	// Set operator scope to the namespace where the operator pod exists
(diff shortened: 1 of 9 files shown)


zolug:
```
cat <<EOF | kubectl apply -f -
apiVersion: meridio.nordix.org/v1
kind: Conduit
metadata:
  name: load-balancer-b1
  namespace: default
  annotations:
    update-sync-group: "kiscsoport"
    resource-template: "small"
  labels:
    trench: trench-a
spec:
  type: stateless-lb
---
apiVersion: meridio.nordix.org/v1
kind: Conduit
metadata:
  name: load-balancer-a1
  namespace: default
  annotations:
    update-sync-group: "kiscsoport"
    resource-template: "medium"
  labels:
    trench: trench-a
spec:
  type: stateless-lb
EOF
```

PR #560, opened 2025-04-09
Proxy to skip reserved linux kernel routing table IDs

zolug:
## Description
Do not use reserved linux kernel routing tables for source based routing in proxies.

Polluting IPv6 routing table _local_ with default nexthop routes towards LBs will prevent further successful route add/replace attempts for the particular LBs in other routing tables.

Introduced cleanup of IP rules during Proxy start up (upon first SetVIPs() call) to remove leftover rules after a process/container restart.

Note: In fact, ever since a Conduit CR is tightly coupled with a single Attractor CR there's not much point to maintain a dedicated routing table per VIP address for source based routing. However, in case of a single shared routing table the route cleanup would need to consider other VIPs as well.

## Issue link
NA

## Checklist

- Purpose
    - [x] Bug fix
    - [ ] New functionality
    - [ ] Documentation
    - [ ] Refactoring
    - [ ] CI
- Test
    - [ ] Unit test
    - [ ] E2E Test
    - [x] Tested manually
- Introduce a breaking change
    - [x] No


zolug:
TODO: Check if it's feasible checking generated tableID is already in use. And add printout when tableID gets rotated.

Issue #562, opened 2025-04-11
Possible deadlock in stateless-lb during Stream delete


zolug:
**Describe the bug**
There's a potential deadlock between Stream deletion interfaceMonitor.Unsubscribe()
call and interfaceMonitor.InterfaceCreated/InterfaceDeleted events.

That's because both Steam Delete() and InterfaceCreated/InterfaceDeleted callbacks from interfaceMonitor wants to acquire the Stream's lb.mu mutex. The callbacks would like to do so by calling triggerPendingTargets(). On the other hand, interfaceMonitor also uses a mutex to protect its integrity that is held when notifying clients subscribed to it, while Subscribe/Unsubscribe needs to acquire it as well.

**To Reproduce**
1. Deploy a Trench with a single Conduit, Attractor, Stream, Flow and VIP.
2. Deploy example target application with a single POD that opens the Stream.
3. Enter the linux network namespace of one of the statless-lb PODs. And create a dummy interface with the same name prefix stateless LB care about. Then start a loop that changes the device's state up and down back and forth.
`ip link add name load-balan-DUMB type dummy && ip link set dev load-balan-DUMB up && while [ 1 ]; do ip link set dev load-balan-DUMB down; ip link set dev load-balan-DUMB up; done`
4. Delete the Stream the example target POD has opened.
`kubectl delete Stream stream-a1`
5. Logs of stateless-lb with the dummy interface should show that the Stream delete is not finished as its "Deleted" printout is missing.

**Expected behavior**

**Context**
 - Kubernetes: v1.31
 - Spire: v1.9.4
 - Network Service Mesh: 1.14.2
 - Meridio: master 2055974

**Logs**
Logs with some additional printouts:

```
{"severity":"debug","timestamp":"2025-04-11T13:25:11.953+00:00","service_id":"Meridio-LB","message":"InterfaceCreated","version":"1.0.0","extra_data":{"class":"LoadBalancer","instance"
:"stream-a2","identifierOffset":5145,"interface":"load-balan-FASZ"}}
{"severity":"debug","timestamp":"2025-04-11T13:25:11.953+00:00","service_id":"Meridio-LB","message":"InterfaceCreated","version":"1.0.0","extra_data":{"class":"LoadBalancer","instance"
:"stream-a1","identifierOffset":5000,"interface":"load-balan-FASZ"}}
[...]

PR #563, opened 2025-04-11
lb; prevent deadlock during stream deletion

zolug:
## Description
Fix deadlock between stream deletion interfaceMonitor.Unsubscribe call and interfaceMonitor.InterfaceCreated/InterfaceDeleted events.

Also, in processPendingTargets() update of Forwarding Availability is delayed until retryPendingTargets() has finished before reporting if the Stream had initial Targets yet ended up with no Targets. This could help avoid unnecessary signalling towards the Frontend, e.g. in case links get replaced in short time.

## Issue link
#562 

## Checklist

- Purpose
    - [x] Bug fix
    - [ ] New functionality
    - [ ] Documentation
    - [ ] Refactoring
    - [ ] CI
- Test
    - [ ] Unit test
    - [ ] E2E Test
    - [x] Tested manually
- Introduce a breaking change
    - [x] No

### DIFF
--- a/pkg/kernel/monitor.go
+++ b/pkg/kernel/monitor.go
@@ -1,5 +1,6 @@
 /*
 Copyright (c) 2021 Nordix Foundation
+Copyright (c) 2025 OpenInfra Foundation Europe
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


LionelJouin (approved):
This is from the mistake I made creating the pkg kernel. An interface monitor could be created each time needed instead maybe.
Good catch

PR #566, opened 2025-05-12
Proxy to use single routing table shared by all VIPs and nexthops for source based routing

zolug:
## Description
There's no use case where a single Conduit should be connected to multiple Attractors, thus there's no reason to group nexthop IPs and handle them separately per VIP address. (That is Proxies of a Conduit don't need to connect LBs of multiple Attractors.)
Therefore, use a single tableID shared by all nexthop routes to be referenced by VIP src rules, in order to lookup where to forward outbound packets with VIP src addresses.

some background:
The baseline did not prevent Proxy from using reserved linux kernel routing tables for source based routing.
Yet, for example polluting IPv6 routing table _local_ with default nexthop routes towards LBs would prevent further successful route add/replace attempts for the particular LBs in other routing tables.

Also introduced cleanup of IP rules during Proxy start up (upon first SetVIPs() call) to remove leftover rules after a process/container restart.

## Issue link
NA

## Checklist

- Purpose
    - [x] Bug fix
    - [ ] New functionality
    - [ ] Documentation
    - [ ] Refactoring
    - [ ] CI
- Test
    - [ ] Unit test
    - [ ] E2E Test
    - [x] Tested manually
- Introduce a breaking change
    - [x] No


pkg/kernel/rule.go
```diff
@@ -0,0 +1,82 @@
...
+func FlushRules(ctx context.Context, family int, matchFn RuleMatchFunc) error {
+	rules, err := netlink.RuleList(family)
+	if err != nil {
+		return fmt.Errorf("failed to fetch ip rules: %w", err)
+	}
+
+	logger := log.FromContextOrGlobal(ctx).WithName("FlushRules")
+	for _, rule := range rules {
+		select {
+		case <-ctx.Done():
```
LionelJouin, line 63: Is this really needed? There is nothing blocking / long process in the next part of this function. Or is it really slow to flush the rules? (I saw you put 15 seconds when calling this function)

pkg/kernel/rule.go
```diff
@@ -0,0 +1,82 @@
...
+		return true
+	}
+}
+
+// FlushRules lists and deletes IP rules for the specified family that match the
+// provided RuleMatchFunc.
+// If matchFn is nil, all rules for the given family will be attempted to be deleted.
+// Errors encountered during rule deletion are accumulated and returned.
+func FlushRules(ctx context.Context, family int, matchFn RuleMatchFunc) error {
+	rules, err := netlink.RuleList(family)
```
LionelJouin, line 55: `RuleListFiltered` could be used to now have list all rules. You need to flush only rules with table id to 1 if I understand correctly.

zolug: You are right. I had this change before deciding on using a common routing table.
On the other hand RuleListFiltered also simply dumps all the rules and then applies the filter in any to determine the return set. Yet, I will change the order of the two main commits and use RuleListFiltered as it would be more fitting.

ljkiraly (approved):

PR #567, opened 2025-05-16
lb; minor pendingCh optimizations

zolug:
## Description
No need to have capacity=10 for buffered pending channel. capacity=1 would already can provide the same functionality, and writers should test if the buffer is full instead of blocking in write.

Also, triggerPendingTargets() must release the mutex if stream is not running.

## Issue link
NA

## Checklist

- Purpose
    - [x] Bug fix
    - [ ] New functionality
    - [ ] Documentation
    - [ ] Refactoring
    - [ ] CI
- Test
    - [ ] Unit test
    - [ ] E2E Test
    - [x] Tested manually
- Introduce a breaking change
    - [x] No


bszirtes (approved):
Looks good to me.

Issue #569, opened 2025-05-23
No bridge IP in proxy due to race condition in IPAM


zolug:
**Describe the bug**
There can be a race between multiple IPAM clients trying to allocate let's say a Node range within the same Conduit range.

Currently, the db model does not require the CIDR field to be unique, thus the IPAM server can add more entries with the same CIDR concurrently while serving multiple clients. The IPAM notices the collision right away and each interfering Allocate "thread" will clean-up its offending entry, and will retry with the next CIDR value in the range. If collisions keep occur, then such competing IPAM clients might end up with an error "no more prefix available". That is Allocate() call fails despite not all prefixes or CIDRs in the particular range are taken.

This can lead to a problem when multiple Proxies within the same Conduit try to get a bridge IP around the same time but some could end up without one. Which causes severe problem for collocated TAPAs during their Stream registration (localIPs will be empty due to the lack of extra_prefixes i.e. proxy bridge IP).

IMHO it's more likely to happen if the IPAM server starts up after the proxies, and if there are multiple clients e.g. proxies.

I don't think the db model should allow adding more entries with the same CIDR. Thus, in case of a race, at least 1 of the concurrent clients could allocate the CIDR they are competing for.
AFAIK migration to a new model where CIDR field has uniqueIndex set would fail, which in our case would not be a lethal fault, so the old model would be kept with its legacy behaviour.

**Logs**
```
{"severity":"error","timestamp":"2025-04-23T07:48:37.48+02:00","service_id":"Meridio-proxy","message":"Setting the bridge IP","version":"1.0.0","extra_data":{"class":"Proxy","error":"failed to allocate bridge IP for child name:\"bridge\" subnet:{conduit:{name:\"vpn2-b\" trench:{name:\"sc-diam\"}} node:\"worker-eccd102-c03-pool1-200u-2prd5\"}: rpc error: code = Unknown desc = failed getting node (worker-eccd102-c03-pool1-200u-2prd5) while allocating (bridge): failed to add node (worker-eccd102-c03-pool1-200u-2prd5) in conduit prefix (vpn2-b): no more prefix available"}}
```

PR #571, opened 2025-06-04
Add CIDR uniqueness constraint to sqlite model

zolug:
## Description
Sqlite model is updated to enforce uniqueness based on `parent_id` and `cidr` fields via database constraints,
directly reflecting the IP allocation logic's expectations regarding CIDR uniqueness within the same `parent`.

The model change greatly complements the allocation logic's collision resolution mechanism, and helps in case of concurrent attempts to avoid constant add/delete churn without stable allocation.

It is also expected to improve allocation performance especially when it comes to concurrent attempts.

Note:
If my understanding is correct, in the case of a rollback the newly introduced indexing would be kept and thus the uniqueness constraint would remain.

Bumping the minor version would be preferable considering the ipam sqlite model update.

## Issue link
#569 

## Checklist

- Purpose
    - [x] Bug fix
    - [ ] New functionality
    - [ ] Documentation
    - [ ] Refactoring
    - [ ] CI
- Test
    - [x] Unit test
    - [ ] E2E Test
    - [x] Tested manually
- Introduce a breaking change
    - [x] No

### DIFF
--- a/pkg/ipam/prefix/allocate_test.go
+++ b/pkg/ipam/prefix/allocate_test.go
@@ -19,6 +19,7 @@ package prefix_test
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"sync"
 	"testing"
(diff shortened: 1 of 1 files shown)


bszirtes (approved):
Looks good to me based on our discussion.

ljkiraly (approved):
Looks good.

Issue #580, opened 2025-07-02
IPAM; support release of node prefixes


zolug:
**Is your feature request related to a problem? Please describe.**
In case of worker node scale-in (or failure) the previously associated node prefixes are not released. Thus, for example continuous scale out/in can exhaust the node prefix pool.

**Describe the solution you'd like**
Release associated node prefixes within conduit prefixes after a worker is removed from the system.

**Describe alternatives you've considered**
- Delete then re-add Conduit to force clean-up.
- Manually edit the registry.db file in IPAM.
- Delete then re-add Trench.
- Monitor node availability information via kubernetes API to keep track of valid prefixes.
- Periodically check proxy resources per node e.g., via k8s API to see if any Proxy POD is running. If no PODs are found running for a while on a node, then remove associated prefixes.
- Improve IPAM Garbage Collector: update node prefix as part of each endpoint allocation to reflect last time of usage, periodic refresh of proxy bridge IPs (to ensure steady update of node prefix irrespective of NSM connections), node prefixes must become _expirable_ for the GC logic to include these, adjust migrate logic to change existing node prefixes' _expirable_ attribute

PR #581, opened 2025-07-03
IPAM; Enable garbage collection for node prefixes

zolug:
## Description

- Proxies periodically renew bridge IP allocation with IPAM.
- Node prefixes are now marked as `expirable`, allowing the SQLite storage garbage collector to reclaim stale resources. To prevent active nodes from being prematurely deleted, their `updatedAt` timestamp is periodically refreshed during `IpamServer`'s `Allocate` and `Release` calls. NSM connections are expected to refresh every 2 minutes, securing these periodic node prefix updates (a proxy is expected to maintain connections towards stateless-lb PODs). These updates are also ensured by (the now introduced) periodic bridge IP renewal.
- IPAM's SQLite garbage collector supports recursive delete (to remove descendants of expired items).
- A proxy instance without bridge IP renewal and with no NSM connections to establish can be subject to premature node prefix removal. However, this is not considered a major issue, as such a proxy would already be inoperational before the prefix reclaim and would remain so afterwards.
- It's also not a concern that bridge IP is not marked `expirable`. Firstly, it can be removed by GC's recursive delete logic as part of an expired node prefix. Secondly, bridge IPs are not renewed by proxies running the baseline code; therefore, marking such prefixes `expirable` would lead to their premature removal, especially during scenarios like long upgrades.

TODO:
- Migration of old DB instance.
- "Damping" to avoid excessively frequent updates of node prefixes

## Issue link
#580 

## Checklist

- Purpose
    - [ ] Bug fix
    - [x] New functionality
    - [ ] Documentation
    - [ ] Refactoring
    - [ ] CI
- Test
    - [ ] Unit test
    - [ ] E2E Test
    - [x] Tested manually
- Introduce a breaking change
    - [x] No

### DIFF
--- a/cmd/proxy/internal/config/config.go
+++ b/cmd/proxy/internal/config/config.go
@@ -1,6 +1,6 @@
 /*
 Copyright (c) 2021-2022 Nordix Foundation
-Copyright (c) 2024 OpenInfra Foundation Europe
+Copyright (c) 2024-2025 OpenInfra Foundation Europe
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


bszirtes (approved):
Looks good to me.

ljkiraly (approved):

PR #586, opened 2025-07-25
Make init container configuration conditional in proxy and stateless-lb-frontend

bszirtes:
## Description

Previously, the init container was always configured in the proxy and stateless-lb-frontend, even if not defined in the template. This change adds a check to ensure init container logic only runs when it is explicitly included in the Pod spec, preventing unnecessary init container injection.

## Checklist

- Purpose
    - [X] Bug fix
    - [ ] New functionality
    - [ ] Documentation
    - [ ] Refactoring
    - [ ] CI
- Test
    - [ ] Unit test
    - [ ] E2E Test
    - [X] Tested manually
- Introduce a breaking change
    - [X] No


bszirtes:
\reverify

